PR #86, opened 2015-02-16
Remove base_path from Publishing API request body

jamiecobbett:
Publishing API currently ignores this field and is about to start rejecting it.
It uses the version in the request URL instead.

See: https://github.com/alphagov/content-store/pull/104


404_finder.rb
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env ruby
```
jennyd: Did you mean to include this file in this PR? It would probably be useful to have it in the repo, but it's unrelated to this change.

jamiecobbett: Fixed.

jennyd:
LGTM :+1: :baseball:

PR #88, opened 2015-06-17
Bump Rails to pick up security fixes

issyl0:
- This causes test failures relating to missing secret_token and
  secret_key_base in dev and test. Add them to the correct config
  file. Corresponding alphagov-deployment changes now in https://github.gds/gds/alphagov-deployment/pull/951 and https://github.gds/gds/alphagov-deployment/pull/952.


jennyd:
There was [a change in Rails 4.1.8](https://github.com/rails/rails/compare/v4.1.6...v4.1.11#diff-1846d4882fa63eb5b1271541995cbfdcR37) around where `secret_token` is stored, but I don't see quite why it is now unhappy with having it set in the initializer since it should be falling back to `config/secret_token` if it isn't in `secrets.yml`. It's better to move to the newer convention, though, so let's do this anyway :+1:

PR #91, opened 2015-07-28
Send manual to rummager async

tommyp:
This PR adds `sidekiq` to the App in order to send Manuals to Rummager in the background. I've done this by adding a `SendManualToRummagerWorker` which is triggered from the `PublishingApiManual` class. [Ticket](https://trello.com/c/YUWkNTfN/56-send-hmrc-manuals-to-rummager-in-the-background).


alext:
You'll need a Procfile, and a `config/sidekiq.yml`.  You'll also need the sidekiq initializer.  It should read config details from `config/redis.yml`.  This will need to be overwriten at deploy time to populate the details of the redis server to use.  See collections-publisher if you want an example of this config.

app/workers/send_manual_to_rummager_worker.rb
```diff
@@ -0,0 +1,7 @@
+class SendManualToRummagerWorker
+  include Sidekiq::Worker
+
+  def perform(manual_id, manual_attributes)
+    HMRCManualsAPI.rummager.add_document(FORMAT, manual_id, manual_attributes)
```
alext: The `FORMAT` constant isn't in scope here.

spec/requests/manuals_spec.rb
```diff
@@ -14,8 +14,10 @@
 
     expect(response.status).to eq(200)
     expect(response.headers['Content-Type']).to include('application/json')
+
     assert_publishing_api_put_item('/hmrc-internal-manuals/employment-income-manual', maximal_manual_for_publishing_api)
-    assert_rummager_posted_item(maximal_manual_for_rummager)
+    expect(SendManualToRummagerWorker.jobs.size).to eql(1)
```
alext: If you swap out the sidekiq testing mode for `inline` below, you could leave this test unchanged.  It think that would make it a more complete test.

app/workers/send_manual_to_rummager_worker.rb
```diff
@@ -0,0 +1,7 @@
+class SendManualToRummagerWorker
+  include Sidekiq::Worker
+
+  def perform(manual_id, manual_attributes)
```
jennyd: We should also move the sending of manual sections to Rummager into a worker; a more generic worker which took the format as well as the id and attributes could be used for both manuals and sections.

app/models/publishing_api_manual.rb
```diff
@@ -66,7 +66,7 @@ def save!
     publishing_api_response = HMRCManualsAPI.publishing_api.put_content_item(base_path, to_h)
 
     rummager_manual = RummagerManual.new(base_path, to_h)
-    HMRCManualsAPI.rummager.add_document(FORMAT, rummager_manual.id, rummager_manual.to_h)
+    SendManualToRummagerWorker.perform_async(rummager_manual.id, rummager_manual.to_h)
```
jennyd: How about moving the call to the worker into a `save!` method on `RummagerManual` which is called from here? That would follow the pattern of this `save!` method making the API call, and would mean that the `PublishingAPIManual` doesn't need to know about how the manual gets sent to Rummager.

spec/requests/manuals_spec.rb
```diff
@@ -14,8 +14,10 @@
 
     expect(response.status).to eq(200)
     expect(response.headers['Content-Type']).to include('application/json')
+
     assert_publishing_api_put_item('/hmrc-internal-manuals/employment-income-manual', maximal_manual_for_publishing_api)
-    assert_rummager_posted_item(maximal_manual_for_rummager)
+    expect(SendToRummagerWorker.jobs.size).to eql(4)
```
jennyd: I agree with @alext's [earlier suggestion](https://github.com/alphagov/hmrc-manuals-api/pull/91#discussion_r35665638) of using sidekiq's inline testing mode, which would let us keep the original assertion here (and in the equivalent place for sections). Otherwise we're left without any assertions that manuals and sections are actually sent to Rummager.

Other than that, and needing the corresponding Puppet and alphagov-deployment changes, this looks good :+1:

tommyp:
Updated the tests to use `Sidekiq::Testing.inline!`. /cc @jennyd @alext

spec/spec_helper.rb
```diff
@@ -3,6 +3,11 @@
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
 SimpleCov.start 'rails'
 
+require 'sidekiq/testing'
+
+# https://github.com/mperham/sidekiq/wiki/Testing#testing-worker-queueing-fake
```
jennyd: This comment and the commit message still refer to `fake`...

tommyp: Ah yeah. Good call. Fixed.

PR #92, opened 2015-07-30
Include leading slashes in the link field of manuals sent to Rummager

issyl0:
- The link field to be sent to Rummager should be the full path, but
  currently for manuals it's the slug without the leading slash. This is
  thought to have been copied from Specialist Publisher, and is
  apparently only working by coincidence.
- Trello:
  https://trello.com/c/xhObMTuI/72-all-manuals-are-sent-to-rummager-without-the-leading-slash-on-the-path-in-the-link-field.

### DIFF
--- a/app/models/rummager_manual.rb
+++ b/app/models/rummager_manual.rb
@@ -5,15 +5,14 @@ class RummagerManual < RummagerBase
   end
 
   def id
-    # The id and link are the path without the leading slash
-    strip_leading_slash(@base_path)
+    @base_path
   end
 
   def to_h
     {
       'title'             => @publishing_api_manual['title'],
       'description'       => @publishing_api_manual['description'],
-      'link'              => id,
+      'link'              => @base_path,
       'indexable_content' => nil,
       'organisations'     => [GOVUK_HMRC_SLUG],
       'last_update'       => @publishing_api_manual['public_updated_at'],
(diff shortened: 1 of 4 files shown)


app/models/rummager_manual.rb
```diff
@@ -5,7 +5,7 @@ def initialize(base_path, publishing_api_manual_hash)
   end
 
   def id
-    # The id and link are the path without the leading slash
+    # The id is the path without the leading slash
```
alext: The id and the link should be the same, and both be the full path.

alext:
Some thought needs to be given to how we clean up the existing entries in rummager with the invalid link fields.

app/models/rummager_manual.rb
issyl0: Ah, right. Does https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/rummager_section.rb#L33 need to be unstripped too?

issyl0:
@alext: Good point. I don't know how Rummager works, but something to change all the `link` and `id` fields to have leading slashes?

app/models/rummager_manual.rb
issyl0: Changed this `id` method to return `@base_path` as removing it completely breaks a lot of tests. :confused:

issyl0:
Pushed a WIP commit for the rake task to republish, but I'm getting some errors from ElasticSearch - something about how it doesn't have all the fields it needs. Closing this for now until someone is back on the team who knows about search things, or I can pair with someone who does. (Thanks @tijmenb for your help this morning!)

PR #94, opened 2015-08-06
Include leading slashes in several fields sent to Rummager

issyl0:
- Stop stripping the leading slashes from `id`, `link` and (in the case of manual sections) `manual` fields in the data we send to Rummager. See the commit message for more detail.
- Add Rake tasks to fix the existing data.

Trello: https://trello.com/c/xhObMTuI/72-all-manuals-are-sent-to-rummager-without-the-leading-slash-on-the-path-in-the-link-field.

(Paired with @mikejustdoit on this.)

### DIFF
--- a/app/models/rummager_manual.rb
+++ b/app/models/rummager_manual.rb
@@ -6,8 +6,7 @@ class RummagerManual < RummagerBase
   end
 
   def id
-    # The id and link are the path without the leading slash
-    strip_leading_slash(@base_path)
+    @base_path
   end
 
   def to_h
--- a/app/models/rummager_section.rb
+++ b/app/models/rummager_section.rb
@@ -5,8 +5,7 @@ class RummagerSection < RummagerBase
   end
 
   def id
-    # The id and link are the path without the leading slash
-    strip_leading_slash(@base_path)
+    @base_path
   end
 
   def section_id
--- a/lib/tasks/rummager_republish/fix_leading_slashes.rake
+++ b/lib/tasks/rummager_republish/fix_leading_slashes.rake
@@ -1,0 +1,93 @@
+require 'gds_api/rummager'
+
+namespace :rummager_republish do
+
+  # Use lambdas because methods in Rake tasks are global, which could
+  # cause problems if anyone else were to write more Rake tasks in this
+  # app.
+  base_path = ->(path) { path.starts_with?('/') ? path : "/#{path}" }
+
+  prepare_manual_for_rummager = ->(manual) {
+    manual_data = {
+      'title'             => manual['title'],
+      'description'       => manual['description'],
+      'public_updated_at' => manual['public_timestamp'],
+    }
+    RummagerManual.new(base_path.call(manual['link']), manual_data)
+  }
+
+  prepare_section_for_rummager = ->(section) {
+    details_hash = {
+      'section_id' => section['hmrc_manual_section_id'],
+      'body'       => section['indexable_content'],
+      'manual'     => { base_path: base_path.call(section['manual']) },
+    }
+
+    section_data = {
+      'title'             => section['title'],
+      'description'       => section['description'],
+      'public_updated_at' => section['public_timestamp'],
+      'details'           => details_hash,
+    }
+    RummagerSection.new(base_path.call(section['link']), section_data)
+  }
+
+  desc 'Republish manuals to Rummager with a leading slash. Delete the old ones.'
+  task manuals_with_leading_slash: :environment do
+    rummager = GdsApi::Rummager.new(Plek.current.find('search'))
+    # As of 2015-08-05, there are 5 HMRC manuals. This is a one-off
+    # script to add leading slashes to manuals, and all new manuals
+    # they publish will have leading slashes by default.
+    count = 10
+
+    search_results = rummager.unified_search(filter_format: ['hmrc_manual'],
+                                             count: count.to_s,
+                                             fields: 'title,description,link,indexable_content,last_update')
+
+    raise 'Error: there are more manuals than you are requesting!' if search_results['total'].to_i > count
+
+    all_manuals = search_results['results']
+    broken_manuals = all_manuals.reject { |manual|
+      manual['_id'].starts_with?('/') && manual['link'].starts_with?('/')
+    }
+
+    broken_manuals.each do |manual|
+      rummager.delete_document('hmrc_manual', manual['_id'])
+
+      rummager_manual = prepare_manual_for_rummager.call(manual)
+
+      rummager_manual.save!
+    end
+    puts "Number of manuals fixed: #{broken_manuals.size}."
+  end
+
+  desc 'Republish manual sections to Rummager with a leading slash. Delete the old ones.'
+  task sections_with_leading_slash: :environment do
+    rummager = GdsApi::Rummager.new(Plek.current.find('search'))
+    # As of 2015-08-05, there are 2339 HMRC manual sections. This is a
+    # one-off script to add leading slashes to manual sections, as all
+    # new manuals they publish will have sections with leading slashes
+    # by default.
+    count = 3000
+
+    search_results = rummager.unified_search(filter_format: ['hmrc_manual_section'],
+                                             count: count.to_s,
+                                             fields: 'title,description,link,indexable_content,last_update,hmrc_manual_section_id,manual')
+
+    raise 'Error: there are more manual sections than you are requesting!' if search_results['total'].to_i > count
+
+    all_sections = search_results['results']
+    broken_sections = all_sections.reject { |section|
+      section['_id'].starts_with?('/') && section['link'].starts_with?('/')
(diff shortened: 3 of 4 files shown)


issyl0:
(This follows on from #92 as that was closed and GitHub doesn't cope with force pushing to branches with closed pull requests.)

app/models/rummager_manual.rb
```diff
@@ -6,15 +6,14 @@ def initialize(base_path, publishing_api_manual_hash)
...
-    strip_leading_slash(@base_path)
+    @base_path
   end
 
   def to_h
     {
       'title'             => @publishing_api_manual['title'],
       'description'       => @publishing_api_manual['description'],
-      'link'              => id,
+      'link'              => @base_path,
```
jackscotti: can't this remain `id`?

app/models/rummager_section.rb
```diff
@@ -25,12 +24,12 @@ def to_h
     {
       'title'                   => title,
       'description'             => @publishing_api_section['description'],
-      'link'                    => id,
+      'link'                    => @base_path,
```
jackscotti: Same here.

app/models/rummager_manual.rb
issyl0: I thought it was nicer like this (I had questions like "why is the link the ID?"), but fair enough - as we have an `id` method, we should use it. Changed.

lib/tasks/rummager_republish/fix_leading_slashes.rake
```diff
@@ -0,0 +1,91 @@
...
+    rummager = GdsApi::Rummager.new(Plek.current.find('search'))
+    # As of 2015-08-05, there are 2339 HMRC manual sections. This is a
+    # one-off script to add leading slashes to manual sections, as all
+    # new manuals they publish will have sections with leading slashes
+    # by default.
+    count = 3000
+
+    search_results = rummager.unified_search(filter_format: ['hmrc_manual_section'],
+                                             count: count.to_s,
+                                             fields: 'hmrc_manual_section_id,public_timestamp,title,manual,indexable_content,link')
```
jennyd: We should fetch `description` as well - we [send that field](https://github.com/alphagov/hmrc-manuals-api/blob/d9cc4aaabf1f9e1c69d8818605ce42de09cd35b1/app/models/rummager_section.rb#L27) to Rummager so it might exist for some sections.

`public_timestamp` is `nil` for all the sections in Rummager in Preview. We [send a timestamp as `last_update`](https://github.com/alphagov/hmrc-manuals-api/blob/d9cc4aaabf1f9e1c69d8818605ce42de09cd35b1/app/models/rummager_section.rb#L31), and that field is available from Rummager and populated, so I think we should be fetching that instead of `public_timestamp`.

If the fields were listed in the same order as in `RummagerSection.to_h` it would be a bit easier to compare them to check we haven't omitted anything (except `organisations` and `format` which are constants) - with the changes mentioned above that would be

```
title,description,link,indexable_content,last_update,hmrc_manual_section_id,manual
```

(argh, posted this on the commit first by mistake :( )

lib/tasks/rummager_republish/fix_leading_slashes.rake
```diff
@@ -0,0 +1,91 @@
...
+
+    raise 'Error: there are more manuals than you are requesting!' if search_results['total'].to_i > count
+
+    all_manuals = search_results['results']
+    broken_manuals = all_manuals.reject { |manual|
+      manual['_id'].starts_with?('/') && manual['link'].starts_with?('/')
+    }
+
+    broken_manuals.each do |manual|
+      rummager.delete_document('hmrc_manual', manual['_id'])
```
jennyd: I think it would be safer to delete the old document after creating the new one, in case the script exits unexpectedly halfway through and we have to run it again, having already deleted something without saving the new version (both here and for sections).

lib/tasks/rummager_republish/fix_leading_slashes.rake
```diff
@@ -0,0 +1,91 @@
...
+
+  desc 'Republish manuals to Rummager with a leading slash. Delete the old ones.'
+  task manuals_with_leading_slash: :environment do
+    rummager = GdsApi::Rummager.new(Plek.current.find('search'))
+    # As of 2015-08-05, there are 5 HMRC manuals. This is a one-off
+    # script to add leading slashes to manuals, and all new manual
+    # they publish will have leading slashes by default.
+    count = 10
+
+    search_results = rummager.unified_search(filter_format: ['hmrc_manual'], count: count.to_s)
```
jennyd: We need to define the fields here too because we need `last_update`, not `public_timestamp`, for the same reason as for sections. They would be `title,description,link,indexable_content,last_update` (a subset of the section fields).

lib/tasks/rummager_republish/fix_leading_slashes.rake
jennyd: On the other hand, since the sending to Rummager is now done in the background, that wouldn't actually help... ignore that comment :)

app/models/rummager_section.rb
```diff
@@ -30,7 +29,7 @@ def to_h
       'organisations'           => [GOVUK_HMRC_SLUG],
       'last_update'             => @publishing_api_section['public_updated_at'],
       'hmrc_manual_section_id'  => section_id,
-      'manual'                  => strip_leading_slash(@publishing_api_section['details']['manual']['base_path']),
+      'manual'                  => @publishing_api_section['details']['manual']['base_path'],
```
rboulton: Are you sure you should be changing this to add a leading slash for the "manual" field?  I thought that this field was expected to contain the slug of the manual.

(You will have noticed that this is all a mess - sorry. At some point we may develop schemas for the content held in rummager - but for now the best we have is the "description" metadata in the `field_definitions.json` file in rummager - unfortunately, we've only got that populated for a very few fields, so far, such as https://github.com/alphagov/rummager/blob/master/config/schema/field_definitions.json#L328)

lib/tasks/rummager_republish/fix_leading_slashes.rake
rboulton: Actually, I think it would be better if `public_timestamp` was populated on sections, instead of `last_update`, and `last_update` didn't exist as a field.

The explanation is that, *surprise*, this has got into a bit of a mess.  The two fields have exactly the same semantics (they're meant to hold the date of the last "major" update, or the last update if there's no such thing as a major update for that format).  Basically, they're meant to be the same as the `public_updated_at` field in content-store (https://github.com/alphagov/content-store/blob/55532bfe00c7d4715ff03599185cf221ce4080f9/doc/content_item_fields.md#public_updated_at) - and it would be lovely to rename them to match that name, but we have no

Historically, `public_timestamp` was populated by documents from whitehall, and `last_update` was populated by documents from specialist-publisher.  We noticed the duplication when implementing the rummager schemas, and made specialist-publisher populate both fields, with an aim to remove the `last_update` field at some point - but never got to the point where we're certain that nothing is still using the `last_update` field.

The current status is that `public_timestamp` is populated for nearly all document formats, and `last_update` is populated for only finder documents, and all documents with `last_update` populated also have the same value in `public_timestamp`.

app/models/rummager_section.rb
issyl0: I asked this question in https://github.com/alphagov/hmrc-manuals-api/pull/92#discussion_r35867108 and received no response, so assumed that we needed to be consistent.

lib/tasks/rummager_republish/fix_leading_slashes.rake
issyl0: OK. :-)

lib/tasks/rummager_republish/fix_leading_slashes.rake
rboulton: Specifying exactly the fields you want is good, for efficiency reasons, and because it makes it easier to work out which apps will be affected if a field is changed.  However, see my other comment about thinking you actually want `public_timestamp` here.

lib/tasks/rummager_republish/fix_leading_slashes.rake
issyl0: Thanks Richard, that's interesting. Changing this to use `public_timestamp` instead of `last_update` is a fundamental change and would be out of scope for this story, so we'll write it up and keep it in mind for whenever we next do anything related (I guess). I'll go with Jenny's suggestion for now.

app/models/rummager_section.rb
issyl0: Jenny mentioned that this field might be used to search within a manual. Changing the field to start with a / does indeed make manual searching fail. I'm going to leave it as it was, without a leading slash.

issyl0:
It seems like I've dealt with all the comments here now.

rboulton:
:+1:

issyl0:
Thanks! :sunny:

PR #95, opened 2015-08-07
Send latest change note to Rummager

tommyp:
In order to show change notes on the latest feed for a Topic page we
need to send the latest change note to Rummager. This commit modifies
the `RummagerManual` to send the `latest_change_note` which is part of
the core doctype in Rummager.

Ticket:
https://trello.com/c/kGiiEN7A/68-send-hmrc-manuals-change-notes-to-rummager

### DIFF
--- a/app/models/rummager_manual.rb
+++ b/app/models/rummager_manual.rb
@@ -11,13 +11,14 @@ class RummagerManual < RummagerBase
 
   def to_h
     {
-      'title'             => @publishing_api_manual['title'],
-      'description'       => @publishing_api_manual['description'],
-      'link'              => id,
-      'indexable_content' => nil,
-      'organisations'     => [GOVUK_HMRC_SLUG],
-      'last_update'       => @publishing_api_manual['public_updated_at'],
-      'format'            => MANUAL_FORMAT,
+      'title'              => @publishing_api_manual['title'],
+      'description'        => @publishing_api_manual['description'],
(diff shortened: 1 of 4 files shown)


app/models/rummager_manual.rb
```diff
@@ -19,6 +19,7 @@ def to_h
       'organisations'     => [GOVUK_HMRC_SLUG],
       'last_update'       => @publishing_api_manual['public_updated_at'],
       'format'            => MANUAL_FORMAT,
+      'latest_change_note'=> @publishing_api_manual['details']['change_notes'].first,
```
rboulton: I'm not sure of this, and it's not tested that I can see - is the "change_notes" field definitely in reverse chronological order, such that "first" is latest?

rboulton:
Looks fine to me - just one question that I wasn't sure of (but that is probably fine).

app/models/rummager_manual.rb
tommyp: I couldn't find an example that suggested it either way, as the example in the App only has 1 change note. However the [content item](https://www.gov.uk/api/content/hmrc-internal-manuals/vat-government-and-public-bodies) for a HMRC Manual has the change notes listed with the latest first.

I've updated the tests to have 2 change notes and to check that Rummager only receives the latest.

mikejustdoit:
:+1:

PR #90, opened 2015-07-27
Add mappings of manual slugs to topics in a CSV, and a script to update the CSV

issyl0:
- This is so that we can tag HMRC Manuals to topics, to make them easier for GOV.UK users to find them.
- Trello: https://trello.com/c/P59WEC2z/54-hardcode-the-list-of-hmrc-manuals-slugs-to-topic-and-sub-topic-tags.
- Made the original CSV manually. Sanity checking has been done and I've fixed the typos.
- Paired with Tommy on the script to update it.

PR #97, opened 2015-08-10
DO NOT MERGE Send topics to Publishing API

mikejustdoit:
We're working towards linking HMRC Manuals to topics.

These changes make use of recently added manual slug -> topic content_ids data (https://github.com/alphagov/hmrc-manuals-api/pull/90) and include that information in the data we export to the Publishing API.

The commits are pretty small and the details can be found in their comments.

These changes depend on the related schema changes being merged https://github.com/alphagov/govuk-content-schemas/pull/107

### DIFF
--- a/app/models/publishing_api_manual.rb
+++ b/app/models/publishing_api_manual.rb
@@ -13,10 +13,11 @@ class PublishingAPIManual
 
   attr_reader :slug, :manual
 
-  def initialize(slug, manual_attributes)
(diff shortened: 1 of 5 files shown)


mikejustdoit:
I think it's been agreed that we need to send the topic data with the `details->tags` to get this working, but keep the `links` that we've already put in place since that'll only be needed again later. I assume that mean we need to send slugs in the tags hash as opposed to content_ids

PR #98, opened 2015-08-11
Naming things

jennyd:
- Use correct name for timestamp field in rummager rake tasks: we're fetching the timestamp as `last_update` from Rummager, so need to get this field out of the returned hash to send it back again.
- Stringify key for manual `base_path` in rummager rake task: `RummagerSection#to_h` looks for this key as a string, not a symbol, so we need to fix this for the path to be preserved by the script rather than being set as nil.


issyl0:
Sorry about it not working right away!

PR #100, opened 2015-08-12
Send latest change note as string

tommyp:
With the work completed in 729eff we were incorrectly sending a hash to
Rummager for the latest change note. This commit updates it to send a
string structed like so "Title of section changed - description of
change."

### DIFF
--- a/app/models/rummager_manual.rb
+++ b/app/models/rummager_manual.rb
@@ -18,11 +18,19 @@ class RummagerManual < RummagerBase
       'organisations'      => [GOVUK_HMRC_SLUG],
       'last_update'        => @publishing_api_manual['public_updated_at'],
       'format'             => MANUAL_FORMAT,
-      'latest_change_note' => @publishing_api_manual['details']['change_notes'].first,
+      'latest_change_note' => latest_change_note,
     }
   end
 
   def save!
     SendToRummagerWorker.perform_async(MANUAL_FORMAT, self.id, self.to_h)
+  end
+
+private
+
+  def latest_change_note
+    latest = @publishing_api_manual['details'].fetch('change_notes', []).first
+
+    "#{latest['change_note']} in #{latest['title']}" if latest
   end
 end
--- a/spec/support/rummager_helpers.rb
+++ b/spec/support/rummager_helpers.rb
@@ -8,13 +8,7 @@ module RummagerHelpers
       'organisations'      => ['hm-revenue-customs'],
       'last_update'        => '2014-01-23T00:00:00+01:00',
       'format'             => 'hmrc_manual',
-      'latest_change_note' => {
-        "title" => "Title of a Section that was changed",
-        "section_id" => "ABC567",
-        "change_note" => "Description of changes",
-        "published_at" => "2014-01-23T00:00:00+01:00",
-        "base_path" => "/hmrc-internal-manuals/employment-income-manual/abc567"
-      },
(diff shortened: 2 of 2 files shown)


app/models/rummager_manual.rb
```diff
@@ -18,11 +18,23 @@ def to_h
...
     SendToRummagerWorker.perform_async(MANUAL_FORMAT, self.id, self.to_h)
   end
+
+private
+
+  def latest_change_note
+    latest_change_note = @publishing_api_manual['details']['change_notes'].first
+
+    if latest_change_note
+      "#{latest_change_note['title']} - #{latest_change_note['change_note']}"
```
jennyd: This is what the change note would look like on the topic latest updates page (with another document for comparison):

screen shot 2015-08-12 at 12 54 15

Maybe it should have a longer dash as the separator between the section title and the change note?

tommyp: Added a longer dash.

jennyd: I think that looks better, but it needs changing in the test data as well :)
screen shot 2015-08-12 at 13 06 09

tommyp: That would help.

app/models/rummager_manual.rb
```diff
@@ -18,11 +18,23 @@ def to_h
...
   end
 
   def save!
     SendToRummagerWorker.perform_async(MANUAL_FORMAT, self.id, self.to_h)
   end
+
+private
+
+  def latest_change_note
+    latest_change_note = @publishing_api_manual['details']['change_notes'].first
```
jennyd: We don't require `change_notes` to be present in the request to the API, so we need to handle the field being absent entirely as well as it being an empty array. At the moment this line errors when the request doesn't include `change_notes` with `undefined method first for nil:NilClass`.

app/models/rummager_manual.rb
```diff
@@ -18,11 +18,25 @@ def to_h
...
     SendToRummagerWorker.perform_async(MANUAL_FORMAT, self.id, self.to_h)
   end
+
+private
+
+  def latest_change_note
+    if @publishing_api_manual['details']['change_notes']
+      latest_change_note = @publishing_api_manual['details']['change_notes'].first
+
+      if latest_change_note
```
jennyd: These checks would be clearer as a single condition rather than as nested `if`s:

`if @publishing_api_manual['details']['change_notes'].present?`

app/models/rummager_manual.rb
```diff
@@ -18,11 +18,19 @@ def to_h
...
   def save!
     SendToRummagerWorker.perform_async(MANUAL_FORMAT, self.id, self.to_h)
   end
+
+private
+
+  def latest_change_note
+    latest = @publishing_api_manual['details'].fetch('change_notes', []).first
+
+    "#{latest['title']} — #{latest['change_note']}" if latest
```
jennyd: We've got some feedback on the format of the latest change note from the content team: the em dash isn't appropriate here, and since colons are already used frequently in section titles they aren't a good option either in this context.

We settled on `"#{latest['change_note']} in #{latest['title']}"`for now. That format won't be right for longer full-sentence change notes but [most so far have been short phrases](https://www.gov.uk/hmrc-internal-manuals/vat-government-and-public-bodies/updates) and we can change the format relatively easily when we've seen a few more manuals. HMRC would need to republish the manual to resend the change note to Rummager, but we think they'll be doing that quite often anyway.

issyl0:
:100:

PR #101, opened 2015-08-12
Document how to test publishing in development

jennyd:
I've added this at the bottom of the README (rather than with the request
documentation) because most of the file is documentation of the API for HMRC
rather than for GDS developers.

Also change the level of the tagging headings and fix a space character in another heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The `<slug>` is used as part of the GOV.UK URL for the document.
 [See an example manual](/public/json_examples/requests/employment-income-manual.json)
 
 
-### JSON Schema
+### JSON Schema
 
 [JSON Schema for manuals](public/manual-schema.json)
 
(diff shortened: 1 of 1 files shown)


jennyd:
Please could other people try the curl command out? I used `--data-binary` because I got an encoding error with `--data` but @mikejustdoit gets the error with `--data-binary`...

```
Encoding::UndefinedConversionError in ManualsController#update
"\xE2" from ASCII-8BIT to UTF-8
```

mikejustdoit:
I can reproduce the error using `--data`, but it's raised from a different place.

The `--data` error's trace:
`render json: { status: "error", errors: [message] }, status: 400` (that's line 28 as mentioned below)
```
app/controllers/application_controller.rb:28:in `rescue in parse_request_body'
app/controllers/application_controller.rb:25:in `parse_request_body'
```

mikejustdoit:
And the `--data-binary` error's trace:
`publishing_api_response = HMRCManualsAPI.publishing_api.put_content_item(base_path, to_h)` (<-- this is line 64 as mentioned below)
```
app/models/publishing_api_manual.rb:64:in `save!'
app/controllers/manuals_controller.rb:7:in `update'
```

alext:
It's probably erroring because `public/json_examples/requests/employment-income-manual.json` isn't a valid JSON file (it has comments in it).

dsingleton:
Using `--data-binary` this works for me, but `--data` fails.

Removing the comments from the `json_example` makes `--data` work too. 

@mikejustdoit is going to see if that fixes it for him too. If so we can move the sentiment of the comments somewhere else (`README`, or description in `manual-schema.json`).

mikejustdoit:
Updated and re-provisioned my VM and I started getting 200s back, with and without comments in the JSON :+1:

dsingleton:
:tada:

PR #103, opened 2015-08-13
Run sidekiq by default in development

jennyd:
Sidekiq is now needed for all publishing via the API, so it's useful to have
it running by default in dev.

By adding the web worker to the Procfile and installing foreman, we can run
foreman from the development Procfile so that bowler will run both the rails
server and a sidekiq worker. This follows the pattern of Specialist Publisher
and email alert API.

Related PR: https://github.gds/gds/development/pull/248


mikejustdoit:
:+1:

PR #104, opened 2015-08-19
Topics for manuals

mikejustdoit:
These changes
- load the `lib/manuals_to_topics.csv` into memory
- use that data and data from Content Register to present topics for each Manual

We're planning to allow "tagging" HMRC Manuals to topics. (The changes to see this through aren't included in this pull request.)

To support the current topic-tagging method, a Manual needs a list of its topics' slugs. To support the upcoming new method of topic-tagging, a Manual needs a list of its topics' content IDs.

It was difficult to remember to use the hmrc-manuals-api house style single quotes, please point out anywhere I've used the wrong ones.


spec/lib/manuals_to_topics_loader_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:csv_data) {
+    <<-EOS.strip
+manual slug,topic slugs,topic ids
+#{manual_with_one_topic},topics/topic-a,topic-a
+#{manual_with_multiple_topics},"topics/topic-a,topics/topic-b","topic-a,topic-b"
+    EOS
+  }
+
+  let(:manual_with_one_topic) { "manuals/manual-0" }
+  let(:manual_with_multiple_topics) { "manuals/manual-1" }
```
jennyd: The CSV file has only the slug of the manual, not a path, so it might be clearer to omit the `manuals/` from these examples.

lib/topics.rb
```diff
@@ -0,0 +1,36 @@
...
+
+private
+  attr_reader :manual_slug, :manuals_to_topics, :content_register
+
+  def matching_topics
+    content_ids.map { |content_id|
+      all_available_topics.find { |topic|
+        topic['content_id'] == content_id
+      }
+    }.compact
```
jennyd: This method could be simpler and easier to follow:
```
all_available_topics.select { |topic| content_ids.include? topic['content_id'] }
```

mikejustdoit: Good suggestion... I suppose until the extra API call is a problem, we shouldn't care. But I didn't think an early return was too ugly, anyway, so I put it in while I was there

spec/lib/manuals_to_topics_loader_spec.rb
mikejustdoit: Thanks. Done. I updated the `content_id`s and all, since they do have a specific shape to them

jennyd: Thanks for doing the content_ids too, that looks clearer.

mikejustdoit:
Tried to rearrange the commits so that the gem updates were actually before the need for them. Github didn't seem to like it, so now they're just in a different wrong order :)

spec/lib/manuals_to_topics_loader_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      expect(loader.load.size).to eq(2)
+    end
+
+    context 'manual with one topic' do
+      let(:hash) { loader.load }
+      let(:content_ids) { hash[manual_with_one_topic] }
+
+      it 'has 1 topic content_id' do
+        expect(content_ids.size).to eq(1)
+      end
```
jennyd: I think it's also worth asserting that the content_id is the value we expect - this test would also pass if we were accidentally using the slug column instead of the content_ids from the CSV file.

mikejustdoit: Yes. Done!

lib/topics.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def format_slug(topic)
+    topic['base_path'].sub('/topic/', '')
+  end
+
+  def all_available_topics
+    @all_available_topics ||= content_register.entries('topic').to_a
+
+  rescue GdsApi::HTTPServerError
+    []
```
jennyd: When an existing manual is updated, its representation in both content-store and rummager is overwritten. Rescuing from the content-register being unavailable would mean that the manual would be updated but it would suddenly disappear from the topic pages, and no emails would be sent (even if it had been a major update) because both will depend on the topic slugs being sent. HMRC would have no indication that that had happened.

That sounds sufficiently confusing for everyone that I think it would be better to not publish the manual and return a 503 response to the client. The `ApplicationController` already [handles some upstream errors](https://github.com/alphagov/hmrc-manuals-api/blob/9b982639a43ed01387b7f2a3ef2ea304639dd5dd/app/controllers/application_controller.rb#L11-L21) like this.

jennyd:
The last 3 commits could be squashed into one now (adding the test helper, using it, then moving it into gds-api-adapters). The earlier commit bumping gds-api-adapters could take it up to 24.1.0 so that change wouldn't need to be in the later squashed commit.

lib/topics.rb
mikejustdoit: Thanks @jennyd, I really wasn't thinking clearly about this part of the workflow

mikejustdoit:
@jennyd, I've altered the API Adapters commit as suggested and squashed those commits into b45cd6a (which unfortunately, has made that commit the _latest_ in the commit history)

mikejustdoit:
Oh yeah, "squashing those commits into b45cd6a" involved testing that a `503` from Content Register is propagated

jennyd:
:fireworks:

mikejustdoit:
:cake:

PR #105, opened 2015-08-25
Send topics to Publishing API and Rummager

mikejustdoit:
This pull request continues where https://github.com/alphagov/hmrc-manuals-api/pull/104 left off, actually making use of the data made available by that work.

With these changes, an HMRC Manual's topics are sent to Publishing API as
- slugs in the `tags` field, and
- content IDs in the `links` field

and to Rummager
- as slugs in the `specialist_sectors` field.

The `tags` and `links` have both been added to the HMRC Manual content schema (https://github.com/alphagov/govuk-content-schemas/commit/d170f4031a4c215da91a962dc61ab4abd3db6409 and https://github.com/alphagov/govuk-content-schemas/commit/b869cee746fff67b8b1fc2e6d29f138ab5fb226e).

### DIFF
--- a/spec/requests/manual_sections_spec.rb
+++ b/spec/requests/manual_sections_spec.rb
@@ -41,7 +41,7 @@ describe 'manual sections resource' do
     expect(response.status).to eq(415)
   end
 
-  it 'handles the content store being unavailable' do
+  it 'handles the Publishing API being unavailable' do
     publishing_api_isnt_available
 
     put_json '/hmrc-manuals/employment-income-manual/sections/12345', maximal_section
@@ -49,7 +49,7 @@ describe 'manual sections resource' do
     expect(response.status).to eq(503)
   end
 
-  it 'handles the content store request timing out' do
+  it 'handles the Publishing API request timing out' do
     publishing_api_times_out
 
     put_json '/hmrc-manuals/employment-income-manual/sections/12345', maximal_section
@@ -57,7 +57,7 @@ describe 'manual sections resource' do
     expect(response.status).to eq(503)
   end
 
-  it 'handles some other error with the content store' do
+  it 'handles some other error with the Publishing API' do
     publishing_api_validation_error
 
     put_json '/hmrc-manuals/employment-income-manual/sections/12345', maximal_section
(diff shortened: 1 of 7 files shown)


spec/requests/manual_sections_spec.rb
```diff
@@ -34,33 +34,33 @@
...
-    put '/hmrc-manuals/employment-income-manual/sections/12345', maximal_section.to_json,
+    put "/hmrc-manuals/#{maximal_section_slug}", maximal_section.to_json,
         headers = {'CONTENT_TYPE' => 'text/plain',
                    'HTTP_ACCEPT'  => 'application/json',
                    'HTTP_AUTHORIZATION' => 'Bearer 12345'}
     expect(response.status).to eq(415)
   end
 
-  it 'handles the content store being unavailable' do
+  it 'handles the Publishing API being unavailable' do
```
jennyd: Good to clean these names up :+1:

spec/support/test_data_helpers.rb
```diff
@@ -1,6 +1,54 @@
...
+
+  def maximal_manual_base_path
+    '/hmrc-internal-manuals/employment-income-manual'
+  end
+
+  def maximal_manual_url
+    'https://www.gov.uk/hmrc-internal-manuals/employment-income-manual'
+  end
+
+  def maximal_section_slug
```
jennyd: Elsewhere in the app a section's slug is only the `12345` part, so we should call this something else.

All of the other helper methods you've extracted here are usable in public-facing URLs except for this one (the `/sections` part of the path is used for requests to the API but not for the frontend), even though the helpers are used in tests for requests to the API. Could the name for this make it clear that it's only for use in this context to distinguish it from the others?

jennyd:
Publishing a manual in dev makes it show up on the topic latest updates page :)

screen shot 2015-08-26 at 16 58 19

spec/support/test_data_helpers.rb
mikejustdoit: Oops, @jennyd, you're right. I saw that it was referred to as the Section's ID around the place, but it turns out it's both ID _and_ slug.

I've updated what this method returns and moved the construction of the API endpoint to the tests that use it (https://github.com/alphagov/hmrc-manuals-api/pull/105/files#diff-1adc332363da2dc044d81444339029b1R9)

jennyd:
I deployed the branch to preview, subscribed to email alerts for business-tax/paye, published the example Employment Income Manual (which is a major update) and got an email! :email: :sparkles: :top: :tophat:

PR #106, opened 2015-08-27
Send `public_timestamp` to rummager

tijmenb:
`last_update`  and `public_timestamp` mean the same thing in rummager.

Current behaviour of rummager is that if `public_timestamp` is empty, `last_update` [gets copied into `public_timestamp`](https://github.com/alphagov/rummager/blob/d3637f348612b76ee4de8d2c01048af39ea2a568/lib/indexer/document_preparer.rb#L46).

This commit changes what is sent to rummager so that we populate `public_timestamp` directly and can remove the special field and special code from rummager.

Trello: https://trello.com/c/yY2jFlXl


jennyd:
Are you happy for the `last_update` field to remain populated in Rummager until the currently-published manuals and sections are next published? If so then :+1:

tijmenb:
Yes, that fine. No need to republish anything, this will sort itself out in time.

PR #109, opened 2015-09-17
Stop stripping leading slashes on the `manual` key of sections

issyl0:
- We originally thought in #94 that including the leading slash in the
  `manual` slug data for a manual section would break search within a
  manual so that it wouldn't work at all. Turns out from the quick fix
  Jenny did in alphagov/frontend#874 that this
  breaks search within a manual in a different way. For consistency, and
  the ability to get rid of the temporary workaround, don't prepend
  leading slashes to any path-like fields anymore.
- Fix old sections to have leading slashes on `manual` values in Rummager.
- It was found when testing this Rake task locally that
  `section['title']` is made up of the hardcoded `section_id` and the
  section's title. We were sending the title back to Rummager and
  reprocessing it
  (https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/rummager_section.rb#L15-L17),
  which prepended yet another `section_id` to the title, leading to
  multiple `section_id`s on the beginning of the title. Fix this, too!

(Paired with Jenny on the latter two commits.)

### DIFF
--- a/app/models/rummager_base.rb
+++ b/app/models/rummager_base.rb
@@ -1,7 +1,3 @@
 class RummagerBase
   GOVUK_HMRC_SLUG = 'hm-revenue-customs'
-
-  def strip_leading_slash(path)
-    path.gsub(%r{^/}, '')
-  end
 end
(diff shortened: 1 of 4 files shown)


mikejustdoit:
Good catch on the `"#{section_id} - #{@publishing_api_section['title']}"`

:+1:

PR #110, opened 2015-09-23
Send topics to Publishing API and Rummager (feature-flagged)

jennyd:
- Revert the reversion of the initial merge
- Add a feature flag

No Puppet changes are needed for this to be merged and deployed - publishing of topics is disabled by default in production-like environments if the `PUBLISH_TOPICS` envvar doesn't exist. We'll enable it in Preview only for now.


issyl0:
The Puppet changes are now in https://github.gds/gds/puppet/pull/3431.

PR #112, opened 2015-10-06
Only publish known slugs

davidbasalla:
For https://trello.com/c/8kOVKy90/89-only-allow-publication-of-hmrc-manuals-with-known-slugs-in-production

Reject publishing unknown manual slugs in production environment via use of feature flag.

Concerns:
- In tests we are relying on the actual data, which makes us uncomfortable. Should we look for a solution that would allow us to stub the list of known manual slugs in the tests?
- We've only tested at unit level, should we add a request spec?

### DIFF
--- a/app/models/publishing_api_manual.rb
+++ b/app/models/publishing_api_manual.rb
@@ -10,16 +10,17 @@ class PublishingAPIManual
 
   validates :to_h, no_dangerous_html_in_text_fields: true, if: -> { manual.valid? }
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
-  validates :slug, inclusion: { in: MANUALS_TO_TOPICS.keys, message: "does not match any of the following valid slugs: #{ MANUALS_TO_TOPICS.keys.join(" ") }" }, if: :only_known_hmrc_manual_slugs?
+  validates :slug, inclusion: { in: :known_manual_slugs, message: "does not match any of the following valid slugs: #{ MANUALS_TO_TOPICS.keys.join(" ") }" }, if: :only_known_hmrc_manual_slugs?
   validate :incoming_manual_is_valid
 
-  attr_reader :slug, :manual
+  attr_reader :slug, :manual, :known_manual_slugs
 
   def initialize(slug, manual_attributes, options = {})
     @slug = slug
     @manual_attributes = manual_attributes
     @manual = Manual.new(@manual_attributes)
     @topics = options.fetch(:topics, Topics.new(manual_slug: slug))
+    @known_manual_slugs = options.fetch(:known_manual_slugs, MANUALS_TO_TOPICS.keys)
   end
 
   def to_h
--- a/app/models/publishing_api_manual.rb
+++ b/app/models/publishing_api_manual.rb
@@ -10,7 +10,7 @@ class PublishingAPIManual
 
   validates :to_h, no_dangerous_html_in_text_fields: true, if: -> { manual.valid? }
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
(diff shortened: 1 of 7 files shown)


jennyd:
Nice test coverage of the new functionality, and the feature flag config looks good :flags: 

We should also prevent publishing of sections if they don't belong to a known manual - sections are published via separate requests to the API so that would need to be validated separately.

We've stubbed the manuals-to-topics CSV file in some other tests, eg [topics_spec.rb](https://github.com/alphagov/hmrc-manuals-api/blob/master/spec/lib/topics_spec.rb), but I think we're still using the real data in some other tests because it was going to get more complex than it was worth to avoid it - I remember having a conversation with @mikejustdoit about this but can't remember the exact context. If you have an approach for stubbing that seems reasonable in this context, then go ahead.

app/models/publishing_api_manual.rb
```diff
@@ -10,6 +10,7 @@ class PublishingAPIManual
 
   validates :to_h, no_dangerous_html_in_text_fields: true, if: -> { manual.valid? }
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
+  validates :slug, inclusion: { in: MANUALS_TO_TOPICS.keys }, if: :only_known_hmrc_slugs?
```
jennyd: At the moment the error message returned to the client from this validation isn't very informative from HMRC's point of view: `"Slug is not included in the list"`. Could we customise the message to be more precise about why it's invalid?

You could also add a request spec to test the validation message as we've [done for other validations](https://github.com/alphagov/hmrc-manuals-api/blob/master/spec/requests/validation_spec.rb).

config/initializers/allow_unknown_hmrc_slugs_feature_flag.rb
```diff
@@ -0,0 +1,5 @@
+if Rails.env.production?
+  HMRCManualsAPI::Application.config.allow_unknown_hmrc_slugs = ENV["ALLOW_UNKNOWN_HMRC_SLUGS"].present?
```
jennyd: [Sections also have slugs](https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/publishing_api_section.rb#L19), so could we be more specific in the naming of the envvar and related methods so that it's clear that they refer to manual slugs?

spec/models/publishing_api_manual_spec.rb
```diff
@@ -14,8 +14,10 @@
   end
 
   subject(:publishing_api_manual) {
-    PublishingAPIManual.new('some-slug', attributes, options)
+    PublishingAPIManual.new(slug, attributes, options)
   }
+  let(:slug) { 'some-slug' }
+  let(:attributes) { valid_manual }
```
jennyd: Minor point: now that we've got this line here we don't need the identical ones on lines 28 and 40.

davidbasalla: Addressed this in the commit

config/initializers/allow_unknown_hmrc_slugs_feature_flag.rb
davidbasalla: Addressed this in the commit

app/models/publishing_api_manual.rb
davidbasalla: Updated the message to give full list of valid slugs and add a validation request spec

davidbasalla:
Just pushed the updated branch. Added separate validations for manual slugs for sections and addressed remaining comments, so ready for re-review :)

davidbasalla:
@mikejustdoit @h-lame I've added a custom validator to check for inclusion of slug in known manual slugs. Let me know if you see any issues with this.

davidbasalla:
@mikejustdoit I've just added a request spec for sections to https://github.com/alphagov/hmrc-manuals-api/commit/2b56be7b4821b38a5b034d5012dcc3fd9c8cde45

mikejustdoit:
Thanks @davidbasalla, this is excellent :+1:

PR #114, opened 2015-10-08
Cleanup incorrectly published manuals

h-lame:
This is the code part (Acceptance criteria 2) of https://trello.com/c/Unsoan0D/61-clean-up-test-hmrc-manuals-in-production

The goal is to be able to run a script that will remove a published manual.  Removal here means to push a "gone" object over the manual in publishing-api, ask rummager to delete the manual, and then do the same for any sections that live "under" the manual.  There's no publishing-api method to find a section for a manual, so we issue a rummager search to do this.

The script is a rake task to be invoked as:

    rake remove_hmrc_manuals[slug-to-remove-1,slug-to-remove-2,...,slug-to-remove-n]

Note that because we've used standard rake arguments, you can't put spaces between the commas :(

### DIFF
--- a/app/models/publishing_api_removed_section.rb
+++ b/app/models/publishing_api_removed_section.rb
@@ -34,12 +34,15 @@ class PublishingAPIRemovedSection
     PublishingAPISection.base_path(@manual_slug, @section_slug)
   end
 
+  def base_path_for_rummager
+    base_path.sub(/\A\//,'')
+  end
+
   def save!
     raise ValidationError, "manual section to remove is invalid" unless valid?
     publishing_api_response = HMRCManualsAPI.publishing_api.put_content_item(base_path, to_h)
 
-    # rummager_manual = RummagerSection.new(base_path, to_h)
-    # rummager_manual.save!
+    HMRCManualsAPI.rummager.delete_document(SECTION_FORMAT, base_path_for_rummager)
 
     publishing_api_response
   end
--- a/app/models/publishing_api_removed_section.rb
+++ b/app/models/publishing_api_removed_section.rb
@@ -1,0 +1,41 @@
+require 'active_model'
+require 'valid_slug/pattern'
+
+class PublishingAPIRemovedSection
+  include ActiveModel::Validations
+
+  validates :manual_slug, :section_slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
+
+  attr_reader :manual_slug, :section_slug
+
+  def initialize(manual_slug, section_slug)
+    @manual_slug = manual_slug
+    @section_slug = section_slug
+  end
+
+  def to_h
+    @_to_h ||= {
+      format: 'gone',
+      publishing_app: 'hmrc-manuals-api',
+      update_type: 'major',
+      routes: [
+        { path: base_path, type: :exact },
+      ],
+    }
+  end
+
+  def base_path
+    PublishingAPISection.base_path(@manual_slug, @section_slug)
+  end
+
+  def save!
+    raise ValidationError, "manual section to remove is invalid" unless valid?
+    publishing_api_response = HMRCManualsAPI.publishing_api.put_content_item(base_path, to_h)
+
+    # rummager_manual = RummagerSection.new(base_path, to_h)
+    # rummager_manual.save!
+
+    publishing_api_response
+  end
+
+end
--- a/app/models/publishing_api_removed_section.rb
+++ b/app/models/publishing_api_removed_section.rb
@@ -5,6 +5,10 @@ class PublishingAPIRemovedSection
   include ActiveModel::Validations
 
   validates :manual_slug, :section_slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
+  validates_with InContentStoreValidator,
+    format: SECTION_FORMAT,
+    content_store: HMRCManualsAPI.content_store,
+    unless: -> { errors[:manual_slug].present? || errors[:section_slug].present? }
 
   attr_reader :manual_slug, :section_slug
 
--- a/app/models/publishing_api_removed_section.rb
+++ b/app/models/publishing_api_removed_section.rb
@@ -43,7 +43,7 @@ class PublishingAPIRemovedSection
   end
 
   def save!
-    raise ValidationError, "manual section to remove is invalid" unless valid?
+    raise ValidationError, "manual section to remove is invalid #{errors.full_messages.to_sentence}" unless valid?
     publishing_api_response = HMRCManualsAPI.publishing_api.put_content_item(base_path, to_h)
 
     HMRCManualsAPI.rummager.delete_document(SECTION_FORMAT, base_path_for_rummager)
--- a/app/models/publishing_api_removed_section.rb
+++ b/app/models/publishing_api_removed_section.rb
@@ -8,6 +8,12 @@ class PublishingAPIRemovedSection
 
   attr_reader :manual_slug, :section_slug
 
+  def self.from_rummager_result(rummager_result)
+    raise InvalidJSONError if rummager_result.blank? || rummager_result['link'].blank?
+    slugs = PublishingAPISection.extract_slugs_from_path(rummager_result['link'])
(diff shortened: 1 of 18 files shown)


mikejustdoit:
I can't hold my minor comments in my head forever, so I'm gonna just add them to the lines as I go along. It might be best if you ignore them for now and I let you know when I'm done reviewing so we can determine how important they are. (For instance, I might pick up on a typo now but later negate it by suggesting an alternative implementation for the method it's in, if that makes sense?)

h-lame:
@mikejustdoit no worries.

spec/models/publishing_api_removed_manual_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      expect(updates_path).to eql('/hmrc-internal-manuals/some-manual/updates')
+    end
+  end
+
+  describe '#to_h' do
+    let(:removed_manual) { described_class.new('some-slug') }
+    subject(:removed_manual_as_hash) { removed_manual.to_h }
+
+    it 'is a "gone" format object' do
+      expect(subject[:format]).to eq 'gone'
```
mikejustdoit: Just a minor thing, it'd be good to have brackets here `eq(...)`

spec/models/publishing_api_removed_manual_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    it 'is invalid without a slug' do
+      expect(described_class.new(nil)).not_to be_valid
+    end
+
+    it 'is invalid with a slug that does not match the valid_slug/pattern' do
+      expect(described_class.new("1Som\nSłu9G!")).not_to be_valid
+    end
+  end
+
+  describe '#base_path' do
```
mikejustdoit: Do you think this might be duplicating the tests for `PublishingAPIManual.base_path` as well as testing that method's behaviour?

Would it be over complicating things to test that we make use of `PublishingAPIManual` here instead of comparing the return value?

spec/models/publishing_api_removed_manual_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      expect(base_path).to eql('/hmrc-internal-manuals/some-manual')
+    end
+
+    it 'ensures that it is lowercase' do
+      base_path = described_class.new('Some-Manual').base_path
+      expect(base_path).to eql('/hmrc-internal-manuals/some-manual')
+    end
+  end
+
+  describe '#updates_path' do
```
mikejustdoit: Similar to my comment on `describe '#base_path'`, but for this one I'm not sure `PublishingAPIManual.updates_path` has any existing unit-level coverage. If that's missing, we could add that

app/models/publishing_api_removed_section.rb
```diff
@@ -0,0 +1,54 @@
...
+  def base_path
+    PublishingAPISection.base_path(@manual_slug, @section_slug)
+  end
+
+  def base_path_for_rummager
+    base_path.sub(/\A\//,'')
+  end
+
+  def save!
+    raise ValidationError, "manual to remove is invalid" unless valid?
```
mikejustdoit: Should this error message say "manual section" instead of just "manual"?

spec/models/publishing_api_removed_manual_spec.rb
```diff
@@ -0,0 +1,148 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/publishing_api'
+require 'gds_api/test_helpers/rummager'
+
+describe PublishingAPIRemovedManual do
+  describe 'validations' do
```
mikejustdoit: Maybe it only makes sense for the non-"removed" Manuals' specs (since they have a schema to validate against), but could there be any value in having a valid example as well as the invalid examples for the Removed Manual and Removed Section?

app/models/publishing_api_removed_section.rb
```diff
@@ -0,0 +1,54 @@
...
+require 'valid_slug/pattern'
+
+class PublishingAPIRemovedSection
+  include ActiveModel::Validations
+
+  validates :manual_slug, :section_slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
+
+  attr_reader :manual_slug, :section_slug
+
+  def self.remove!(manual_slug, slug)
```
mikejustdoit: Is this method used?

h-lame: Ah, no.  It's a vestigial tail of a rebased-away approach that didn't quite work.  Will remove.

app/models/publishing_api_removed_section.rb
h-lame: Almost certainly.  Good spot, thanks!

spec/models/publishing_api_removed_manual_spec.rb
h-lame: At the time of writing the specs I wasn't sure I wanted to just re-use `PublishingAPIManual.base_path`, as it happens, I did.  Could rework as a mockist style of test - but at that point I'm not sure the test would be worth much as the spec would be almost the same as the implementation.  How do we feel about that kind of spec in general?

spec/models/publishing_api_removed_manual_spec.rb
h-lame: Yeah, whatever we end up doing for base_path I'll do here.  Either way though, I'll get a spec around `PublishingAPIManual.updates_path`.

spec/models/publishing_api_removed_manual_spec.rb
h-lame: I guess that goes for all the `eq`s? I'll update them.

spec/models/publishing_api_removed_manual_spec.rb
h-lame: Ah, well, there's no specs for `PublishingAPIManual#base_path` or `PublishingAPIManual#update_path` just the class versions.  Maybe I'll get rid of these for consistency.  How would we feel about that as a solution?

mikejustdoit: Yeah, getting rid sounds good to me

spec/models/publishing_api_removed_manual_spec.rb
mikejustdoit: Thanks

spec/models/publishing_api_removed_manual_spec.rb
mikejustdoit: Ha, sorry, I didn't word that so well, did I?

Yep all the `eq`s. And I think everything else non-`eq` is already written that way

spec/models/publishing_api_removed_manual_spec.rb
h-lame: I'd usually put one of these in if we had a factory, although my purpose with that is to check that the factory generates a useful object, rather than provide a "it is valid when..." spec for the object itself.

h-lame:
@mikejustdoit Addressed your comments so far in rebased commits (apart from the "valid" spec one).  Keep 'em coming ;)

spec/models/publishing_api_removed_manual_spec.rb
mikejustdoit: Yep, that makes sense

mikejustdoit:
I've tried this out on my dev VM and it mostly works excellently. I've just noticed one _small_ thing we'll need to address before it can be merged, sorry.

We don't currently check whether a manual with the specified slug exists. The rake task could be a little bit more informative to the user about this, but more importantly, we shouldn't publish "gone" content items for manuals that didn't exist in the first place.

Sorry to not've spotted this sooner :disappointed:

mikejustdoit:
If that's a pain to figure out, just shout if you want to pair on it

h-lame:
@mikejustdoit - yeah, that makes sense.  Might need a pointer on how to ask `<whatever>` if the manual exists.  Doesn't feel like `Rummager` is the right place to ask, although I think it'd be simple.

h-lame:
Latest 3 commits ensure that a removed manual or section "exists" before allowing it to be removed. /cc @mikejustdoit

mikejustdoit:
Excellent work @h-lame :+1:

PR #116, opened 2015-10-19
Include base_path's leading slash in request

mikejustdoit:
Rummager identifies HMRC Manuals and Sections by their path including the leading slash.

It looks weird, but Rummager requests with these identifiers in their paths should include the leading slash and therefore contain a double slash ("//").

### DIFF
--- a/app/models/publishing_api_removed_manual.rb
+++ b/app/models/publishing_api_removed_manual.rb
@@ -44,10 +44,6 @@ class PublishingAPIRemovedManual
     PublishingAPIManual.base_path(@slug)
   end
 
-  def base_path_for_rummager
-    base_path.sub(/\A\//,'')
-  end
-
   def updates_path
     PublishingAPIManual.updates_path(@slug)
   end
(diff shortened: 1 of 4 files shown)


mikejustdoit:
While I was writing the commit message, it didn't seem useful to describe the specific problem we were having here, but maybe I should?

rboulton:
@mikejustdoit Describing the problem in the commit message would have been useful - it would still be useful to add it to this ticket.

mikejustdoit:
OK, here's the problem as I see it. I'm not sure what I'm describing is particularly useful:

We're using Rummager's `delete_document`, but until now the documents weren't being deleted as expected.

HMRC Manuals are indexed by Rummager with an ID copied from their `base_path` in this app. In stripping the leading slash from our Manual's `base_path`, we were sending Rummager a document ID it didn't recognise, and therefore couldn't delete.

From where we were looking, requests that "looked fine" were being made to Rummager, but the documents were still appearing in the search results.

PR #120, opened 2015-11-18
Explain content_id in README

tijmenb:
This commit adds the content_id to the JSON examples and attempts to explain the content_id in the readme.

Some of it is lifted from https://github.com/alphagov/content-store/blob/master/doc/content_item_fields.md#content_id

/cc @rboulton

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,27 @@ Additionally, users of the API are required to follow the styleguide for slugs:
 
 Section IDs are validated to ensure that they can be converted to slugs by simply making them lowercase.
 
+## Content IDs
+
+Both manuals and sections have a field `content_id`, which will be required in
+the future.
+
+This is a UUID string as described in [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) ([Wiki](https://en.wikipedia.org/wiki/Universally_unique_identifier)). It is
+[validated using this regex](https://github.com/alphagov/publishing-api/blob/1bd2c3d2aaa4681fe6286548aa16a4b5f66367c9/app/validators/uuid_validator.rb#L10-L24).
+
+For example: "30737dba-17f1-49b4-aff8-6dd4bff7fdca".
+
+This is a unique identifier for the piece of content. It is used as the reference
+with which content items can reference other content items on GOV.UK. For example,
+it is used for [tagging to topics](https://www.gov.uk/topic).
+
+Manuals and sections should always have a consistent `content_id`. It is not possible
+to send a previously-published document with the same slug but a different `content_id`.
+
+Currently, when a manual or section doesn't have a `content_id`, one will be
+generated for it. This generated UUID is non-random and based on the `base_path`
+of the item.
+
 ## Markup
 
 All `body` attributes in manuals or manual sections may contain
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -106,6 +106,19 @@ Additionally, users of the API are required to follow the styleguide for slugs:
 
 Section IDs are validated to ensure that they can be converted to slugs by simply making them lowercase.
 
+## Content IDs
+
+Both manuals and sections have a required field `content_id`.
```
jennyd: This field isn't required yet.

README.md
```diff
@@ -106,6 +106,19 @@ Additionally, users of the API are required to follow the styleguide for slugs:
...
+
+This is a UUID string as described in [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) ([Wiki](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
+
+For example: "30737dba-17f1-49b4-aff8-6dd4bff7fdca".
```
jennyd: It might be useful to include a link to the [regex in the Publishing API](https://github.com/alphagov/publishing-api/blob/1bd2c3d2aaa4681fe6286548aa16a4b5f66367c9/app/validators/uuid_validator.rb#L10-L24) that this will be validated against, so that it's clear exactly what we will accept (lower-case, hyphenated).

README.md
```diff
@@ -106,6 +106,19 @@ Additionally, users of the API are required to follow the styleguide for slugs:
...
+
+This is a unique identifier for the piece of content. It is used as the reference with which content items can reference other content items on GOV.UK. For example, it is used for [tagging to topics](https://www.gov.uk/topic).
+
+Manuals and sections should always have the same content_id. It is not possible
```
jennyd: This sounds as though sections should have the same `content_id` as the manual they belong to, rather than sections and manuals having unique and consistent content ids over time.

tijmenb:
@jennyd updated PR. We should probably wait for https://github.com/alphagov/hmrc-manuals-api/pull/122 to be merged tho

jennyd:
Now that https://github.com/alphagov/hmrc-manuals-api/pull/122 is merged, I think we should explain it here so that this special behaviour is clearly documented, both for us and for HMRC.

README.md
```diff
@@ -106,6 +106,21 @@ Additionally, users of the API are required to follow the styleguide for slugs:
...
+the future.
+
+This is a UUID string as described in [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) ([Wiki](https://en.wikipedia.org/wiki/Universally_unique_identifier)). It is
+[validated using this regex](https://github.com/alphagov/publishing-api/blob/1bd2c3d2aaa4681fe6286548aa16a4b5f66367c9/app/validators/uuid_validator.rb#L10-L24).
+
+For example: "30737dba-17f1-49b4-aff8-6dd4bff7fdca".
+
+This is a unique identifier for the piece of content. It is used as the reference with which content items can reference other content items on GOV.UK. For example, it is used for [tagging to topics](https://www.gov.uk/topic).
+
+Manuals and sections should always have a consistent content_id. It is not possible
```
jennyd: Minor point: non-backticked content_id.

tijmenb:
@jennyd 

I've added the following:

> Currently, when a manual or section doesn't have a `content_id`, one will be
> generated for it. This generated UUID is non-random and based on the `base_path`
> of the item.

Does that make sense?

jennyd:
Yup :+1:

PR #121, opened 2015-11-20
Generate content id when absent

steventux:
This is a rough proposal for working around the missing `content_id` values for `hrmc_manual` and `hmrc_manual_section` formats.
Full story here https://trello.com/c/pULByuHs/393-fix-invalid-content-store-data
The uuid is generated consistently from the `base_path` of the manual/section, it's a valid v5 uuid as per RFC4122. It's also discernible from a typical `SecureRandom.uuid` as these are v4.

/cc @jennyd Can we talk about this when you have a moment.

### DIFF
--- a/app/models/helpers/publishing_api_helpers.rb
+++ b/app/models/helpers/publishing_api_helpers.rb
@@ -12,5 +12,19 @@ module Helpers
       )
       attributes
     end
+
+    def add_absent_content_id(attributes)
+      unless attributes["content_id"]
+        attributes["content_id"] = base_path_uuid
+      end
+
+      attributes
+    end
+
+    private
+
+    def base_path_uuid
+      UUIDTools::UUID.sha1_create(UUIDTools::UUID_URL_NAMESPACE, base_path).to_s
+    end
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,8 @@ gem 'govspeak', '3.0.0'
 
 gem 'sidekiq', '3.4.2'
 
+gem 'uuidtools', '2.1.5'
+
 group :development do
   gem "foreman", "0.78.0"
 end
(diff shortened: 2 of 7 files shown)


tijmenb:
Recap from irl conversation, @rboulton and I approve of this. We will ask the HMRC devs to use the content-ids generated with this method for existing content. :+1:

jamiecobbett:
I don't think this would handle slugs being changed - though if I recall correctly, hmrc-manuals-api doesn't support that anyway?

jennyd:
This seems like a reasonable approach to unblock the import :+1: 

The API doesn't currently support slugs being changed. It is possible that manual slugs might need to change in future, which this approach wouldn't support. If a manual slug changes then the base path for all its sections will also change.

We could tell HMRC that we're taking this approach now to unblock the import, and that it will prevent them from changing manual slugs in future (after they've been published) unless they are first able to start sending content ids.

We'll also need to add the content ids into the `gone` representations of [manuals](https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/publishing_api_removed_manual.rb) and [sections](https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/publishing_api_removed_section.rb).

app/models/helpers/publishing_api_helpers.rb
```diff
@@ -12,5 +12,19 @@ def add_organisations_to_details(attributes)
...
+
+    def add_absent_content_id(attributes)
+      unless attributes["content_id"]
+        attributes["content_id"] = base_path_uuid
+      end
+
+      attributes
+    end
+
+    private
```
jennyd: Minor style point: could you [outdent `private`](https://github.com/alphagov/styleguides/blob/master/ruby.md#classes)?

steventux:
Thanks all for comments and review, moved this PR to https://github.com/alphagov/hmrc-manuals-api/pull/122

PR #123, opened 2015-12-03
Migrate to v2 of publishing-api

stephen-richards:
All apps should now use v2 of the publishing api to write content to the content-store. This change converts hmrc-manuals-api to use v2 of the api. The differences are:

- put content_item is replaced by three methods - put_content, publish, and put_links
- the content_id is used as the key, rather than the base path
- the base path is included in the payload, links are not included in the payload of the put_content

The three requests to publishing-api have been refactored out into a PublishingAPINotifier class which halts processing if any of the responses are non-200, and returns the response to the client.

Trello: https://trello.com/c/II6tydnO/453-use-publishing-api-v2-for-hmrc-manuals-api


tijmenb:
https://github.com/alphagov/hmrc-manuals-api/blob/859a43e6e06919651195161c0c4ae1d42455d1bb/app/models/publishing_api_manual.rb#L118-L125 also adds links to the content-item sent to the publishing-api. These need to go separately with a `put_links` request.

rboulton:
Regarding the links thing - perhaps we shouldn't be setting this, and should just enable content tagger for hmrc-manuals-api content.

tijmenb:
Offline recap: this app will start using content-tagger for tagging soon, but that is a larger piece of work (Trello: https://trello.com/c/6Owhx9Vn). For now we need to keep the functionality and use `put_links` to set the topics.

tijmenb:
@stephen-richards https://github.com/alphagov/govuk-content-schemas/pull/164 will update the content-schemas to remove the `links` from the main content-item.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -0,0 +1,31 @@
+class PublishingAPINotifier
+
+  def initialize(document)
+    @document = document
+  end
+
+  def notify
```
tijmenb: This probably doesn't do what we want yet. 

If the HTTP calls for `put_content`, `publish`, `put_links` return 4XX or 5XX, [gds-api-adapters will raise an error](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api_v2.rb), not return the status code.

Also, the methods we're calling here (`put_content`, `publish`, `put_links`) could (for example) also return `201 Created`, which should succeed.

spec/models/publishing_api_manual_spec.rb
```diff
@@ -106,6 +106,21 @@
     end
   end
 
+  describe 'set_previous_version' do
+    let(:manual)  { PublishingAPIManual.new(slug, attributes, options) }
+    it 'should populate the hash if it hasnt already been populated' do
```
tijmenb: Try to avoid `it "should ..."`, since it will cause every test to start this way. In this case `it 'populates the hash...'` is better.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -0,0 +1,30 @@
+class PublishingAPINotifier
+  def initialize(document)
+    @document = document
+  end
+
+  def notify
+    response = put_content_item
+    update_document_version(response)
```
tijmenb: Sending the version number can perhaps be simplified.

Something like:

```ruby
content_item = put_content_item
publish(previous_version: content_item.version)
```

and then:

```ruby
def publish(attrs = {})
  HMRCManualsAPI.publishing_api.publish(@document.content_id, @document.update_type, attrs)
end
```


This means you can get rid of the `previous_version` attributes, `update_document_version` and all the tests,

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,39 @@
+module PublishingAPIStubHelpers
```
tijmenb: There are a lot of new methods here that may not be necessary. Could you see if you can use the methods from the gem for some cases? https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L13

tijmenb:
There's still a `.DS_Store` file in this PR, could you remove that?

spec/support/publishing_api_stub_helpers.rb
stephen-richards: These are all convenience methods to save a lot of repeated code that would be necessary using the methods from the gem.  In some cases we want to control what is returned, and others what is sent.

tijmenb:
There are still a few references to `previous_version` in the code, which aren't used anymore.

app/models/publishing_api_removed_manual.rb
```diff
@@ -31,10 +31,11 @@ def sections_from_rummager
 
   def to_h
     @_to_h ||= {
-      content_id: base_path_uuid,
+      base_path: base_path,
```
tijmenb: This still has `content_id`, which you may not want.

If you change `publisher` to `publisher_v2` [here](https://github.com/alphagov/hmrc-manuals-api/blob/c30cfe91381984b49010a07f4bc0c3f3b77386c9/spec/support/govuk_content_schemas.rb#L4), we'll test against the new schemas and be sure we send the right stuff.

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+  def publishing_api_endpoint
+    GdsApi::TestHelpers::PublishingApiV2::PUBLISHING_API_V2_ENDPOINT
+  end
+
+  def stub_publishing_api_publish_with_version(content_id, version, update_type = 'major')
```
tijmenb: You could get rid of this by making `stub_publishing_api_publish(content_id, body)` accept a `response`:

```
def stub_publishing_api_publish(content_id, body, response_body)
  #...
end
```

https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L21

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+    stub_publishing_api_publish(content_id, {update_type: update_type, previous_version: version}.to_json)
+  end
+
+  def stub_publishing_api_put_content_publish_and_put_links(content_id)
```
tijmenb: This is almost:

https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L31

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+    stub_publishing_api_put_content_and_return_response(content_id, {version: 22}.to_json)
+    stub_publishing_api_publish_with_version(content_id, 22)
+    stub_any_publishing_api_put_links
+  end
+
+  def stub_publishing_api_put_content_and_return_response(content_id, response)
```
tijmenb: Would be useful to make `stub_publishing_api_put_content` return a response too: 

https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L13

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+    url = "#{publishing_api_endpoint}/content/#{content_id}"
+    stub_request(:put, url).to_return(status: 200, body: response, headers: {"Content-Type" => "application/json; charset=utf-8"})
+  end
+
+  def stub_publishing_api_put_content_returning(status)
+    stub_request(:put, /#{publishing_api_endpoint}\/content\/.*/).to_return(status: status)
+  end
+
+  def publishing_api_isnt_available
```
tijmenb: This would be useful to add to the api helpers too.

spec/support/publishing_api_stub_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+    stub_request(:any, /#{publishing_api_endpoint}\/.*/).to_return(status: 503)
+  end
+
+  def assert_publishing_api_put_links(content_id, attributes_or_matcher = {}, times = 1)
```
tijmenb: This is also very useful in the test helpers.

spec/support/publishing_api_stub_helpers.rb
tijmenb: Yes, but I think they could better live in the `gds-api-adapters` test helpers, so that other people can use them and we don't duplicate across projects. Most of them seem useful to others as well.

app/models/publishing_api_removed_manual.rb
stephen-richards: removed now.

app/notifiers/publishing_api_notifier.rb
stephen-richards: Noted, and fixed.

app/models/publishing_api_manual.rb
```diff
@@ -90,15 +101,27 @@ def topic_slugs
...
+    HMRCManualsAPI::Application.config.publish_topics && topic_content_ids.present?
+  end
+
+  def topic_links
+    { links: { topics: topic_content_ids } }
+  end
+
 private
+  def generate_content_id_if_absent
+    if @manual_attributes.is_a?(Hash)
```
tijmenb: When is it not a hash?

spec/support/publishing_api_data_helpers.rb
```diff
@@ -77,10 +84,10 @@ def maximal_section_for_publishing_api(options = {})
     {
       "format" => "hmrc_manual_section",
       "title" => "A section on a part of employment income",
-      "content_id" => "25e687b8-74da-4892-938d-7de82fa5df27",
+      # "content_id" => "25e687b8-74da-4892-938d-7de82fa5df27",
```
tijmenb: Could you remove the commented out code?

spec/requests/validation_spec.rb
```diff
@@ -56,7 +56,9 @@
       end
 
       it 'allows images with a relative path' do
-        stub_default_publishing_api_put
+        content_id = UUIDTools::UUID.sha1_create(UUIDTools::UUID_URL_NAMESPACE, "/hmrc-internal-manuals/imaginary-slug").to_s
```
tijmenb: I think `stub_any_publishing_api_call` is more appropriate here.

tijmenb: Nevermind last comment, this is because of the versioning.

app/models/publishing_api_manual.rb
stephen-richards: When the payload is invalid - see spec/models/publishing_api_manual_spec.rb:117.

There is a validation rule that checks for this, but is not called until later.  We could call it before generating the content_id, as part of the initializer I guess.

spec/support/publishing_api_data_helpers.rb
```diff
@@ -125,18 +130,19 @@ def maximal_section_for_publishing_api(options = {})
...
   end
 
-  def gone_manual_for_publishing_api(slug: 'some-slug')
+  def gone_manual_for_publishing_api(base_path: '/hmrc-internal-manuals/some-slug')
     {
+      'base_path' => base_path,
       'format' => 'gone',
       'publishing_app' => 'hmrc-manuals-api',
-      'update_type' => 'major',
+      # 'update_type' => 'major',
```
jackscotti: you have missed this commented out line

spec/support/publishing_api_data_helpers.rb
```diff
@@ -147,7 +153,7 @@ def gone_manual_section_for_publishing_api(manual_slug: 'some-manual', section_s
     {
       'format' => 'gone',
       'publishing_app' => 'hmrc-manuals-api',
-      'update_type' => 'major',
+      # 'update_type' => 'major',
```
jackscotti: you have missed this commented out line

jackscotti:
looks good to me, but would like to get @tijmenb to merge this

PR #124, opened 2015-12-11
Remove all sections when removing manual

jennyd:
Rummager defaults to returning 10 results if a count isn't specified, so the
rake task to remove a manual and all its sections was only removing up to 10
sections and leaving the rest. The largest manuals we know of have less than
3000 sections, so a limit of 5000 should always be large enough.

Also withdraw the manual after its sections, so that if the script exits early for some reason it won't error on a second run.


davidbasalla:
Looks good to me :+1:

PR #126, opened 2016-01-13
Resolve issue with autoloading `API::UserController`

benlovell:
Prior to this change, we were experiencing problems with this
application inflecting `Api` to `API` and thus failing to autoload the
`gds-sso` supplied `API::UserController`.

Removing the inflection resolves this issue and the tests remain green.
I did spend some time trying to ascertain whether inflecting `API` was
important and didn't find any further evidence within this application.

cc @jennyd, @benilovj

### DIFF
--- a/config/initializers/acronym.rb
+++ b/config/initializers/acronym.rb
@@ -1,5 +1,4 @@
 ActiveSupport::Inflector.inflections do |inflect|
   inflect.acronym 'HTML'
-  inflect.acronym 'API'
   inflect.acronym 'HMRC'
 end


benilovj:
:+1: from me but I'll leave it up to @jennyd to decide if this is an acceptable solution to lose this inflection.

h-lame:
Can we get some more context on this?  On the face of it it seems legit as the `API` acronym inflection probably would interfere with loading `Api::UserController`, but as there are plenty of models in the app with a file name like `foo_api_bar` that contain classes with names like `FooAPIBar` so I'd be worried that we end up with the same problem in reverse.

How did we trigger the original problem?  FWIW: with or without the inflection I can't make `Api::UserController` or any of the `FooAPIBar` models complain - but rails autoloading is totally weird so it wouldn't surprise me to find out that there is a problem.

benlovell:
@h-lame there are four models containing `API` and they're loaded correctly since the constants are cased without needing help from inflections. Nowhere are they accessed in the form `Api` and I'd rather this didn't linger if possible.

h-lame:
Can replicate the error and this definitely seems to fix it.  It also doesn't seem to cause any errors with autoloading our `API` classes.  I think I'd be more comfortable if we did the work to invert the commit that added this inflection in the first place, 9612e0bb5a15fb565559191d1777ebbfeb087b9b, bearing in mind there are more than the two models in that original commit.

benlovell:
@h-lame that sounds reasonable. Do you folks want to pick this up from here?

h-lame:
I've done some more digging and convinced myself that we're going to be ok here.  It seems that the `AS::Inflector` will cope with an `API` class living in an `api` filename without our custom inflections, but won't cope with an `Api` class in an `api` filename with our custom inflections.  Which explains the problem.

I reckon were done here.  Merging.

benlovell:
Thanks :+1:

PR #127, opened 2016-01-15
Use `Services` module for external APIs

mobaig:
This app is using an old pattern to retain references to the API
clients. The `Services` module is simpler and is in use in most other
applications.


benlovell:
Nice :sparkler:

jennyd:
Which other apps are using this pattern? I don't remember seeing it in many others on GOV.UK. If we're going to standardise on a pattern like this it's worth discussing a bit more widely.

PR #128, opened 2016-01-19
Remove all code related to writing topic slugs and links

mobaig:
Tagging of hmrc manuals will be handled by the content tagger app. This
entire subset of functionality dealing with topic/manual mappings and
the writing of topic data to the details/links hashes is no longer
required in this codebase.

As the publishing of topic data is currently switched off in production, this should be safe to merge and deploy even in the absence of the required config changes to content-tagger that enable tagging for hmrc manuals (although those changes will be going in across the next few days).

Trello: https://trello.com/c/6Owhx9Vn


app/models/publishing_api_manual.rb
```diff
@@ -102,15 +87,8 @@ def save!
...
-  def send_topic_links?
-    HMRCManualsAPI::Application.config.publish_topics && topic_content_ids.present?
-  end
-
-  def topic_links
-    { links: { topics: topic_content_ids } }
-  end
-
 private
+
```
benlovell: This is quickly becoming my de facto PR comment: our style guide says methods should immediately follow access modifiers. It might be worth hooking this project up with `govuk-lint` to catch and correct the usual linting misses.

rboulton:
This code looks correct to me.  I'd merge, but I want the "custom" team to have a look over it first.

config/initializers/known_manual_slugs.rb
```diff
@@ -0,0 +1,222 @@
...
+  vat-womens-sanitary-products
+  vat-youth-clubs
+  venture-capital-schemes-manual
+  video-games-development-company-manual
+  wine-manual
+  holding-and-movements-alcohol-stategy-abandonment-of-goods
+  holding-and-movements-assurance-guidance
+  hydrocarbon-oils-strategy
+  scottish-taxpayer-technical-guidance
+}
```
Davidslv: You might need to freeze this constant to avoid any changes

app/models/publishing_api_manual.rb
mobaig: Gah, genuinely thought styleguide stated leave a space after. Will chuck this through the linter and commit some cleanup.

config/initializers/known_manual_slugs.rb
mobaig: Seems sensible, will update.

app/models/publishing_api_manual.rb
mobaig: Hmmm...the linter catches a bunch of other stuff but doesn't actually detect the spacing around private.

config/initializers/known_manual_slugs.rb
h-lame: It's no bad thing to freeze the array, but it's worth pointing out that it doesn't give you total protection against changes here.  In this case, we can't add or remove elements from `KNOWN_MANUAL_SLUGS`, nor can we directly swap out elements, but we can still edit in place the individual elements:

```ruby
a = ["hats", "cats"].freeze # => ["hats", "cats"]
a << 'mats' # => RuntimeError: can't modify frozen Array
a[0] = 'mats' # => RuntimeError: can't modify frozen Array
a[0] << 'mats' # => "hatsmats"
a # => ["hatsmats", "cats"]
```

This is because only the array is frozen, not the elements within it.  To get total protection we could freeze the individual elements too:

```ruby
a.each { |elem| elem.freeze } # => ["hatsmats", "cats"]
a[1] << 'mats' # => RuntimeError: can't modify frozen String
```

I wouldn't actually suggest we do this though because it's unusual and anyone encountering it will need to take a moment to work out why before moving on.

h-lame:
This seems legit.  Do we need to talk to HMRC about this change?  It was off in production so I'm guessing not.  I'll check before I merge though.

issyl0:
But... but... the CSV I spent two days manually compiling and then fixing my typos! :sob: :stuck_out_tongue_winking_eye:

app/controllers/manuals_controller.rb
```diff
@@ -6,9 +6,11 @@ def update
...
       publishing_api_response = manual.save!
       respond_to do |format|
-        format.json { render json: { govuk_url: manual.govuk_url },
-                    status: publishing_api_response.code,
-                    location: manual.govuk_url }
+        format.json {
+          render json: { govuk_url: manual.govuk_url },
+          status: publishing_api_response.code,
+          location: manual.govuk_url
+        }
```
jamiecobbett: I found this block a little hard to parse. Indenting lines 12 and 13 to make it clearer that they are arguments to `render` would help.

h-lame:
After discussion with @marksheldonGDS we'll merge this after the content-tagger changes are live.

app/controllers/manuals_controller.rb
mobaig: Done.

mobaig:
@h-lame @marksheldonGDS The content-tagger change is now live. HMRC manuals can be tagged up with topics and organisation via the content-tagger web interface.

As discussed, making topic updates via content-tagger will not currently result in email updates being sent out. A conversation earlier today with @h-lame suggests this may not be an issue. Further, I'll very shortly be working on adding support for email notifications on updates to a document's links hash.

Naturally @marksheldonGDS should confirm first, but I think this code change can now be merged.

rboulton:
Having discussed this with @jennyd I'm happy to merge this.

PR #129, opened 2016-01-25
Conditionally add organisation to links hash of incoming manual updates

mobaig:
When handling an incoming payload containing updates for a manual or
manual section, check if it already has an organisation specified in the
content store. If not, set the organisation to HMRC by default.

This change includes an update to the PublishingApiNotifier that allows client
code to specify whether a put_links update should take place as part of
the notify. This is useful when removing manuals/sections, a case when no links
update is required.

Trello: https://trello.com/c/6Owhx9Vn

### DIFF
--- a/app/models/links_builder.rb
+++ b/app/models/links_builder.rb
@@ -1,0 +1,28 @@
+class LinksBuilder
+  def initialize(content_id)
+    @content_id = content_id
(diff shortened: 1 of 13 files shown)


rboulton:
This looks good to rebase and merge once #128 is merged.

PR #133, opened 2016-02-26
Update README

jennyd:
- Update references to Preview environment
- Fix rendering of a code block

I think that everything else in the README is still accurate.


brenetic:
:+1:

PR #134, opened 2016-03-07
2 new slugs added and amendments to 2 other slugs

markmcleod:
HMRC have requested we add 2 new slugs to this list of agreed ones. They have also requested a change to 2 other slugs. Checked with GDS content team.


davidbasalla:
Looks good, merging...

PR #135, opened 2016-03-22
Changes to list of HMRC manual slugs

markmcleod:
For https://trello.com/c/pCJdcZXl/349-review-changes-to-hmrc-manual-slugs


config/initializers/known_manual_slugs.rb
```diff
@@ -96,7 +96,7 @@
   lloyds-manual
   machine-games-duty-manual
   mlr1-penalties-guidance
```
h-lame: Based on the change below it seems odd that they don't want to change this from `mlr1` to `money-laundering` - should we ask them?

h-lame: For closure: they do want to make this rename, but this manual is already published so we need to do something more than this to handle it.  Other PRs will address that.

PR #137, opened 2016-04-07
Update README following styleguide

davidbasalla:
We move the current README contents to a file in the 'docs' folder and create a new README that is more consistent with our [style guide](https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md). We moved the original detailed README content to the '/docs' folder and link to appropriate header tags from the new README.

We want to keep serving the original content under the '/readme' route of the app for users of the app, so we update the 'readme' controller method to return the content from the new file location.

For https://trello.com/c/0X2x4mr3/357-review-and-update-readme-files-for-4-of-our-applications-3

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -8,16 +8,16 @@ class RootController < ActionController::API
     render :index
   end
 
-  def readme
-    content = load_readme
+  def documentation
+    content = load_documentation
     content = rewrite_public_links(content)
-    render :readme, locals: { content: content }, layout: true
+    render :documentation, locals: { content: content }, layout: true
   end
 
 private
 
-  def load_readme
-    file_contents = File.open(Rails.root + "README.md", 'r').read
+  def load_documentation
+    file_contents = File.open(Rails.root + "docs/extended_documentation.md", 'r').read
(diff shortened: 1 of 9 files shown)


app/controllers/root_controller.rb
```diff
@@ -17,7 +17,7 @@ def readme
 private
 
   def load_readme
```
jennyd: Could we rename the methods here, the route and the [link on the front page](https://github.com/alphagov/hmrc-manuals-api/blob/28940bad525d5c0e26c9bf7fc00230fec4590461/app/views/root/index.html.erb#L4) to `documentation` instead of `readme` for consistency? We should redirect `/readme` to `/documentation` in case anyone has bookmarked that URL too.

davidbasalla: I've updated the PR with another commit to handle the redirect.

davidbasalla:
I've added a LICENCE file which was missing previously. I copied it from https://github.com/alphagov/contacts-admin/blob/update-readme/LICENCE

emmabeynon:
I've proof-read this and it looks fine

sihugh:
:+1:

PR #139, opened 2016-04-21
Amended 2 slugs and added 7 new slugs

markmcleod:
Amended 2 slugs for unpublished manuals.

Added 7 new slugs for unpublished manuals.


davidbasalla:
Looks good to me, there are no previous entries for the amended slugs, so will merge 👍

PR #140, opened 2016-04-25
Update link paths in docs for JSON schemas

davidbasalla:
The paths were not updated when the extended documentation was moved to the docs folder in https://github.com/alphagov/hmrc-manuals-api/commit/a439160d10cab98197da9a25adf1449e9a66048c, and therefore would 404.


brenetic:
👍

PR #142, opened 2016-04-26
Add new production assets hostname

alexmuller:
We're going to be moving assets to this hostname at some point.

I've left the old one in because there's no reason to remove it now.


jennyd:
Could you add this domain to the [docs about images](https://github.com/alphagov/hmrc-manuals-api/blob/0a544d4691be16bee897c7bfab57c88f260df106/docs/extended_documentation.md#images) too?

alexmuller:
Have done, thanks! How does that sound?

jennyd:
👍

PR #144, opened 2016-05-19
Remove and redirect sections

davidbasalla:
We need to remove a parent section and some of its subsections, and redirect to a new parent section.
We add 2 rake tasks, one for posting 'gone' items for sections and one for posting 'redirect' items. 
When we run this in production, we intend to first run the 'remove-sections' rake task to remove the required subsections. After that we run the 'redirect-section' rake task to redirect from the old parent section to the new parent section.

In the last commit, we add the shoulda-matchers gem to shorten our validation tests. In order for the matchers to work, we need to make the instance variables available with the `attr_accessor` call. See commit notes for details.

For https://trello.com/c/HdwQ2pIw/354-unpublish-and-redirect-a-section-of-a-hmrc-manual

Paired with @sihugh and @klssmith


h-lame:
I'm a little bit down on the `ActiveModel::Model` thing.  The reduction in the spec verbosity is nice, but I'm not convinced by the increase in API that we grant the `PublishingAPIRedirectedSection` in order to achieve it (`ActiveModel::Model` brings a lot more than validations and hash-based initializers).  Does `shoulda-matchers` only work with `ActiveModel::Model` instances or can we get the same benefit with a smaller change (e.g. writing our own hash-based initializer?).

Either way it would be nice to go fully commit to the approach and apply whatever changes we make to `PublishingAPIRedirectedSection` to the other similar objects and their specs too.

davidbasalla:
@h-lame thanks! I haven't had any luck with a hash-based initializer yet, but will give it another try.
Agreed, if we can find a solution, we'll roll it out to the other models as well.

davidbasalla:
@h-lame We've updated the PR. Turns out I completely overreacted with including ActiveModel::Model. In the end, we only had to change `attr_reader` to `attr_accessor` for the instance variables that we validate and set the subject in rspec to have sensible defaults.

h-lame:
@davidbasalla: Oh, that's much neater 🍰

app/models/publishing_api_redirected_section.rb
```diff
@@ -0,0 +1,63 @@
...
+      ],
+    }
+  end
+
+  def content_id
+    base_path_uuid
+  end
+
+  def redirect_to_location
+    File.join(PublishingAPIManual::BASE_PATH_SEGMENT, destination_manual_slug, destination_section_slug)
```
davidbasalla: I think this could be `PublishingAPISection.base_path(destination_manual_slug, destination_section_slug)`...

h-lame: Looks like it, yeah.

h-lame:
Looks good!

benilovj:
Not to put a spanner in the works, but I think that we should be using the newly created [unpublishing endpoint of the Publishing API](https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#post-v2contentcontent_idunpublish) for this. It does a fair amount of work for you, and better long-term for the stack, in my opinion.

/cc @elliotcm

h-lame:
@benilovj - oh! didn't even know that existed.  In the interest of getting this PR done so we respond to the support ticket, would you be ok if we merged this as-is and raised a ticket / PR to convert the system over in total to the new API at a later date?  There's a bunch of things that do "gone" vs. "redirect" publishings in here, so reworking to use the new API could be a lot of work.

benilovj:
@h-lame that makes sense, but it's best not to leave that work for too long, lest it be forgotten...

h-lame:
@benilovj - we've written it up in our trello as https://trello.com/c/v994rYfL/262-update-hmrc-manuals-api-to-use-unpublishing-endpoint-from-publishing-api - feel free to review and let us know if we should make changes.

benilovj:
@h-lame thanks

PR #147, opened 2016-06-07
Add a rake task to add the HMRC organisation to a subset of manual sections

mobaig:
3124 HMRC manual sections have been identified by the Finding Things
migration checker as missing their organisation link. This is a simple,
temporary rake task to patch HMRC into the links of these sections.


tijmenb:
Note that these 3124 are manual sections that have been not been republished after this application started sending organisations properly to the publishing-api (https://github.com/alphagov/hmrc-manuals-api/pull/129). Normally, we would have triggered a mass-republish after such a change, but because HMRC is in charge of this data we can't do that. A data migration to reconcile the data is therefore the only solution.

PR #149, opened 2016-06-20
Remove organisation link from Rummager payload

mobaig:
Merge when we've switched on Rummager's publishing API fetch.

Why?:

Rummager has recently been updated to fetch links data directly from the
publishing API when indexing a document. Setting Rummager links data in
the publishing app is therefore redundant.

How?:

- Strip organisations from the payload in RummagerManual.
- Strip organisations from the payload in RummagerSection.

### DIFF
--- a/app/models/rummager_manual.rb
+++ b/app/models/rummager_manual.rb
@@ -16,7 +16,6 @@ class RummagerManual < RummagerBase
       'description'        => @publishing_api_manual['description'],
       'link'               => id,
       'indexable_content'  => nil,
-      'organisations'      => [GOVUK_HMRC_SLUG],
       'public_timestamp'   => @publishing_api_manual['public_updated_at'],
       'format'             => MANUAL_FORMAT,
       'latest_change_note' => latest_change_note,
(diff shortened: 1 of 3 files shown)


tijmenb:
I think there's also:

https://github.com/alphagov/hmrc-manuals-api/blob/abe7c0a9d8d0eb36c4662689b220cdd16182c6ca/app/models/rummager_section.rb

mobaig:
@tijmenb yep, jumped the gun on the PR. Need some test fixes too.

tijmenb:
The constant can be removed too I think.

mobaig:
@tijmenb The constant is still used here: https://github.com/alphagov/hmrc-manuals-api/blob/master/app/models/rummager_section.rb#L46

Filtering by organisation seems superfluous if it's already filtering on format and manual_path. Unless I'm missing something `'filter_organisations' => [GOVUK_HMRC_SLUG]` can be removed. WDYT?

tijmenb:
👍 Good to merge once we've deployed https://trello.com/c/0FKSqAkK

Not sure what is going on with that filtering.

PR #151, opened 2016-08-12
Upgrade Rails to 4.2.7.1

emmabeynon:
This upgrade is necessary due to a possible XSS Vulnerability in Action View

Advisory: CVE-2016-6316
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/I-VWr034ouk

and an Unsafe Query Generation Risk in Active Record

Advisory: CVE-2016-6317
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/rgO20zYW33s

The upgrade from Rails 4.1 to Rails 4.2 necessitated the inclusion of the `responders` gem to handle respond_to methods (see: http://guides.rubyonrails.org/v4.2/upgrading_ruby_on_rails.html#responders) and removal of the rails-api gem.


h-lame:
Be good to also update any config files that differ between rails 4.2 and 4.1.  http://railsdiff.org/4.1.0/4.2.7.1 can be used to check what the differences are.  This means future upgrades will be easier as we'll already have modern settings and such.

emmabeynon:
Thanks @h-lame - I've updated the existing config files and added missing ones where necessary.

h-lame:
:+1:

PR #152, opened 2016-08-19
Add script to check for orphaned sections for a manual or section

brenetic:
We receive requests from HMRC for the removal of manuals or sections.
Before we do this it is necessary to check whether any sections that
belong to the thing being removed has been given another parent section
or manual. This avoids any sections becoming orphans.

HMRC should be able to do this themselves before raising a request with
us. This script checks for any sections that have not been re assigned
to another parent and lists those that require changing.

[Trello card](https://trello.com/c/cIESLI1q/463-write-a-script-to-automate-whether-an-hmrc-manuals-section-can-be-deleted-3)

cc @h-lame

### DIFF
--- a/lib/sections_checker.rb
+++ b/lib/sections_checker.rb
@@ -1,0 +1,96 @@
+class SectionsChecker
+  def initialize(item_slug)
+    @item_slug = item_slug
+  end
+
+  def check
+    item = content_item(item_base_path)
+
+    children = child_sections(item)
+
+    if item['format'] == MANUAL_FORMAT
+      check_children_of_manual(item, children)
+    elsif item['format'] == SECTION_FORMAT
+      check_children_of_section(item, children)
+    end
+  end
+
+private
+
+  def item_base_path
+    PublishingAPIManual.base_path(@item_slug)
+  end
+
+  def content_item(base_path)
+    Services.content_store.content_item(base_path)
+  end
+
+  def child_sections(item)
+    item['details']['child_section_groups'].flat_map do |group|
+      group['child_sections']
+    end
+  end
+
+  def check_children_of_manual(manual, children)
+    find_children_of_parent_manual(children, manual).map do |child|
+      child['base_path']
+    end
+  end
+
+  def find_children_of_parent_manual(children, parent_manual)
+    children.select do |child|
+      child_to_check = content_item(child['base_path'])
+      belongs_to_parent_manual?(child_to_check, parent_manual)
+    end
+  end
+
+  def belongs_to_parent_manual?(child, parent_manual)
+    if child['details']['manual']['base_path'] == parent_manual['base_path']
+      true
+    else
+      parent = content_item(child['details']['manual']['base_path'])
+      !child_section_in_tree?(parent, child)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


lib/sections_checker.rb
```diff
@@ -0,0 +1,95 @@
+class SectionsChecker
```
davidbasalla: Naming quibble... would `OrphanedSectionsChecker` be clearer in terms of the class' responsibility?

brenetic: Möglicherweise, however we are checking all the sections and sections of sections in order to find a potential orphan . . .

davidbasalla:
Looking at the test cases for when given a section slug, would it be useful to add another test for a section with grand children? The point being that it would exercise the recursive behaviour of the code as presumably it will need to walk down the full length of a branch. As far as I can see, right now we're just testing the immediate children of a section... happy to discuss

lib/sections_checker.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def check
+    item = content_item(item_base_path)
+
+    children = child_sections(item)
+
+    if item['format'] == MANUAL_FORMAT
+      check_children_of_manual(item, children)
+    elsif item['format'] == SECTION_FORMAT
+      check_children_of_section(item, children)
```
davidbasalla: Is it possible that we're missing some sort of recursive lookup when we check children of a Section? Tracing down the code path of `check_children_of_manual(item, children)` eventually calls the recursive method `child_section_in_tree?`. When tracing down the code path for `check_children_of_section(item, children)` no recursive function gets called... is that correct?

davidbasalla:
Looks good, can't think of any edge cases that aren't covered! 👍  Took quite a while to review, in part because of the recursive tree structure, but also the way that parent/child relationships are defined in this app/content-store. I imagine the whole thing would be easier if the associations were handled in an ActiveRecord like manner, so that one could rely on something like `manual.children`/`section.parent` calls.

lib/sections_checker.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def belongs_to_parent_manual?(child, parent_manual)
+    if child['details']['manual']['base_path'] == parent_manual['base_path']
+      true
+    else
+      !child_section_in_tree?(child['details']['manual']['base_path'], child)
+    end
+  end
+
+  def child_section_in_tree?(base_path, child)
```
davidbasalla: Was wondering whether this could take the parameter list `(parent, child)`, so taking the parent content item rather than its `basepath` so that the two parameters are at the same level of abstraction and are easier to understand at first glance. 

That could mean that we update Line 51 to something like:
```
other_manual = content_item(child['details']['manual']['base_path'])
!child_section_in_tree?(other_manual, child)
```

Feel free to ignore if you don't agree, just wondering how to make this more readable for future us, as it took me a while to read through.

h-lame: This seems like a sensible suggestion.

brenetic:
@davidbasalla  Thanks for the comments. Have updated.

davidbasalla:
Merging, great stuff 👍

PR #155, opened 2016-09-01
Fix Rummager request limit

sihugh:
Since April 2016 [Rummager has set a limit of 1000 records](https://github.com/alphagov/rummager/commit/8f93aac19164f30ab5e46e190a6825c23f6ab35a).
The present query uses 5000, which provokes the error message `Maximum result
set size (as specified in 'count') is 1000` to be returned.


tijmenb:
Sorry about this, I think we missed this project when adding that limit.

sihugh:
No probs, thanks for the merge 👍

jennyd:
Just to record this here, I'm fairly sure that some manuals have over 1000 sections (maybe 2-3000?) which is why we set the count so high in the first place. We may need to make several requests to find all the sections to remove when we're withdrawing a whole manual (or can we get the full list from the publishing API now instead?).

PR #157, opened 2016-09-13
Add Lint to default rake task and fix deprecation warnings

sihugh:
Added lint to the default rake task and fixed the problems.

Also fixes the deprecation warnings:

> DEPRECATION NOTICE: You are making requests that will potentially return nil. Please set `GdsApi.config.always_raise_for_not_found = true` to make sure all responses with 404 or 410 raise an exception. Raising exceptions will be the default behaviour from October 1st, 2016.

and

> stub_any_rummager_post_with_queueing_enabled is deprecated: use stub_any_rummager_post instead

from gds-api-adapters

And corrected the spelling of a private method.

### DIFF
--- a/app/models/links_builder.rb
+++ b/app/models/links_builder.rb
@@ -14,12 +14,13 @@ class LinksBuilder
   end
 
 private
+
   def set_organisation
-    if @content_store_links["organisations"].present?
-      @built_links["organisations"] = @content_store_links["organisations"]
-    else
-      # Use HMRC content ID to set organisation
-      @built_links["organisations"] = ["6667cce2-e809-4e21-ae09-cb0bdc1ddda3"]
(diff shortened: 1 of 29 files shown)


app/models/links_builder.rb
```diff
@@ -9,17 +9,18 @@ def build_links
...
-  def set_organistion
-    if @content_store_links["organisations"].present?
-      @built_links["organisations"] = @content_store_links["organisations"]
-    else
+
+  def set_organisation
+    @built_links["organisations"] = if @content_store_links["organisations"].present?
+                                      @content_store_links["organisations"]
+                                    else
       # Use HMRC content ID to set organisation
```
issyl0: Can you indent this comment too?

PR #159, opened 2016-09-22
Update gds-api-adapters behaviour

emmabeynon:
[Trello card](https://trello.com/c/B5z6Dq8q/)

Some old behaviour has been deprecated in [gds-api-adapters](https://github.com/alphagov/gds-api-adapters) so we are updating this app to comply with this before it becomes the default. Firstly, a GdsApi::NotFound error is now raised when the server returns a 404 or 410.  Secondly, requests now return a hash instead of an OpenStruct.  We have now opted in to this new behaviour.


app/models/links_builder.rb
```diff
@@ -5,18 +5,18 @@ def initialize(content_id)
   end
 
   def build_links
-    @content_store_links = Services.publishing_api
-      .get_links(@content_id)
-      .try(:links)
-      .to_h.with_indifferent_access
+    @content_store_link_set = Services.publishing_api.get_links(@content_id)
```
h-lame: With the new behaviour of gds-api-adapters that we're opting into I don't think this API call will ever return `nil`.  The old behaviour was to return `nil` if there's a 404/410 whereas the new behaviour is to raise an exception.

emmabeynon: Thanks @h-lame - I've modified a test that was allowing nil to be returned, and updated the code accordingly.  Let me know what you think.

spec/support/links_update_helper.rb
```diff
@@ -10,4 +10,9 @@ def stub_put_default_organisation(content_id)
       { links: { organisations: ["6667cce2-e809-4e21-ae09-cb0bdc1ddda3"] } }.to_json
     )
   end
+
+  def stub_publishing_api_get_no_links(content_id)
+    stub_request(:get, Plek.new.find('publishing-api') + "/v2/links/#{content_id}")
+      .to_raise(GdsApi::HTTPNotFound)
+  end
```
h-lame: We could replace this with `publishing_api_does_not_have_links ` (see:  https://github.com/alphagov/gds-api-adapters/blob/059f83d7b5f01549cd527501104983cf1bd3fc20/lib/gds_api/test_helpers/publishing_api_v2.rb#L244-L247).  Using the test helpers from the gems should mean our tests are more robust in the face of future changes in the behaviour of the adapters.

app/models/links_builder.rb
```diff
@@ -5,18 +5,19 @@ def initialize(content_id)
...
   def build_links
-    @content_store_links = Services.publishing_api
-      .get_links(@content_id)
-      .try(:links)
-      .to_h.with_indifferent_access
+    begin
+      @content_store_links = Services.publishing_api.get_links(@content_id)["links"].with_indifferent_access
+    rescue GdsApi::HTTPNotFound
+      @content_store_links = nil
+    end
```
h-lame: :+1:

app/models/links_builder.rb
h-lame: I don't know if I can mark this comment as resolved somehow, but it is.

spec/support/links_update_helper.rb
emmabeynon: Thanks @h-lame - this should now be changed.

h-lame:
Good stuff.

PR #160, opened 2016-09-28
Add tasks to redirect manual sections to their parent manuals

sihugh:
Added the ability to page over Rummager results, as we know that there are manuals
 with over 1000 sections (and at least one with > 5000)

Added two rake tasks: `redirect_hmrc_section_to_parent_manual` and
`redirect_all_sections_to_parent_manual`.  At present we don't have any need
to redirect sections to other manuals or manuals to other manuals.

The `redirect_all_sections_to_parent_manual' task outputs the source and
destination slugs which could be handy if something went awry.

Added a new model `PublishingAPIRedirectedSectionToParentManual` in the same
vein as `PublishingAPIRedirectedSection`.  There seems to be enough difference
in the validations (for example) to keep it separate.

Added a `SectionRetriever` class to encapsulate the construction and execution
of Rummager queries, and allow construction of result items as required so we
can construct both the `Redirected` and `Removed` section types from it.
I tried shoehorning it into `PublishingAPIManual`, but it didn't seem to fit
very nicely.  We don't need it for anything that PublishingAPIManual does at
the moment, so it seemed wrong to add complexity there.

Also removed some duplication from the remove/redirect rake tasks

https://trello.com/c/Wu6gbYDj/464-unpublish-pages-from-hmrc-manuals

### DIFF
--- a/app/models/publishing_api_removed_manual.rb
+++ b/app/models/publishing_api_removed_manual.rb
@@ -24,10 +24,35 @@ class PublishingAPIRemovedManual
   end
 
   def sections_from_rummager
-    query = RummagerSection.search_query(base_path)
-    Services.rummager.search(query)["results"]
+    sections = []
+    search_response = nil
+
+    loop do
+      new_query = rummager_section_query(start_index(search_response))
+
+      search_response = Services.rummager.search(new_query)
+      sections += search_response["results"]
+      return sections if all_sections_retrieved?(sections, search_response)
+    end
   end
+
   private :sections_from_rummager
+
+  def all_sections_retrieved?(sections, search_response)
+    sections.count >= search_response["total"]
+  end
+
+  def start_index(search_response)
+    if search_response
+      search_response["start"] + search_response["results"].count
+    else
+      0
+    end
+  end
+
+  def rummager_section_query(start_index)
+    RummagerSection.search_query(base_path, start_index)
+  end
 
   def to_h
     @_to_h ||= {
--- a/app/models/publishing_api_removed_manual.rb
+++ b/app/models/publishing_api_removed_manual.rb
@@ -18,40 +18,9 @@ class PublishingAPIRemovedManual
   end
 
   def sections
-    sections_from_rummager.map do |section_json|
+    SectionRetriever.new(slug).sections_from_rummager.map do |section_json|
       PublishingAPIRemovedSection.from_rummager_result(section_json)
     end
-  end
-
-  def sections_from_rummager
-    sections = []
-    search_response = nil
-
-    loop do
-      new_query = rummager_section_query(start_index(search_response))
-
-      search_response = Services.rummager.search(new_query)
-      sections += search_response["results"]
-      return sections if all_sections_retrieved?(sections, search_response)
-    end
-  end
-
-  private :sections_from_rummager
-
-  def all_sections_retrieved?(sections, search_response)
-    sections.count >= search_response["total"]
-  end
-
-  def start_index(search_response)
-    if search_response
-      search_response["start"] + search_response["results"].count
-    else
-      0
-    end
-  end
-
-  def rummager_section_query(start_index)
-    RummagerSection.search_query(base_path, start_index)
   end
 
   def to_h
--- a/lib/section_retriever.rb
+++ b/lib/section_retriever.rb
@@ -1,0 +1,42 @@
+class SectionRetriever
+  attr_reader :manual_slug
+
+  def initialize(manual_slug)
+    @manual_slug = manual_slug
+  end
+
+  def sections_from_rummager
+    sections = []
+    search_response = nil
+
+    loop do
+      new_query = rummager_section_query(start_index(search_response))
+
+      search_response = Services.rummager.search(new_query)
+      sections += search_response["results"]
+      return sections if all_sections_retrieved?(sections, search_response)
+    end
+  end
+
+private
+
+  def all_sections_retrieved?(sections, search_response)
+    sections.count >= search_response["total"]
+  end
+
+  def start_index(search_response)
+    if search_response
+      search_response["start"] + search_response["results"].count
+    else
+      0
+    end
+  end
+
+  def rummager_section_query(start_index)
+    RummagerSection.search_query(base_path, start_index)
+  end
+
+  def base_path
+    PublishingAPIManual.base_path(manual_slug)
+  end
+end
--- a/lib/tasks/redirect_sections.rake
+++ b/lib/tasks/redirect_sections.rake
@@ -1,22 +1,5 @@
 desc "Redirect a section, takes original manual slug, section slug, destination manual slug and destination section slug"
 task :redirect_hmrc_section, [] => :environment do |_task, args|
-  def output_error_message(message)
-    puts "ERROR!"
-    print "  "
-    puts message
-  end
-
-  def redirect_and_output(manual)
-    response = manual.save!
-    if response.code == 200
-      puts "OK!"
-    else
-      output_error_message(response.raw_response_body)
-    end
-  rescue => e
-    output_error_message(e.message)
-  end
-
   slugs = args.extras
   if slugs.empty? || slugs.length < 4
     puts "Usage: rake redirect_hmrc_section[manual-slug,section-slug-to-redirect,destination-manual-slug,destination-section-slug]"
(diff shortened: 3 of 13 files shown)


app/models/publishing_api_removed_manual.rb
```diff
@@ -24,11 +24,32 @@ def sections
   end
 
   def sections_from_rummager
-    query = RummagerSection.search_query(base_path)
-    Services.rummager.search(query)["results"]
+    results = []
```
davidbasalla: What do you think about renaming `results` to `sections`? It might make it a bit clearer what sort of data we're dealing with.

Also we could wrap the final condition into a method called `all_sections_retrieved?(sections, search_response)`. That might make this method clearer to understand at first glance.

app/models/publishing_api_removed_manual.rb
```diff
@@ -24,11 +24,32 @@ def sections
...
 
+  def start_index(search_response)
+    if search_response
+      search_response["start"] + search_response["results"].count
+    else
+      0
+    end
+  end
+
+  def query(start)
```
davidbasalla: The method `query` could be renamed to `rummager_section_query` to be more explicit. And maybe rename the input parameter from `start` to `start_index`?

lib/tasks/redirect_sections.rake
```diff
@@ -30,3 +13,52 @@ task :redirect_hmrc_section, [] => :environment do |_task, args|
     redirect_and_output(section)
   end
 end
+
+desc "Redirect a section to its parent manual, takes original manual slug, section slug, destination manual slug"
+task :redirect_hmrc_section_to_parent_manual, [] => :environment do |_task, args|
```
davidbasalla: `redirect_hmrc_section_to_parent_manual` could be renamed to `redirect_section_to_parent_manual`, as the hmrc is assumed?

davidbasalla: Actually, looking at the other rake tasks, looks like they have the word `hmrc` in them as well. In that case, maybe rename `redirect_all_sections_to_parent_manual` to `redirect_all_hmrc_sections_to_parent_manual` so that it matches the pattern of the other tasks

lib/tasks/redirect_sections.rake
```diff
@@ -30,3 +13,52 @@ task :redirect_hmrc_section, [] => :environment do |_task, args|
...
+  slugs = args.extras
+  if slugs.empty? || slugs.length < 3
```
davidbasalla: Should this be `slugs.length < 2`, as it only requires two arguments?

app/models/publishing_api_redirected_sections_to_parent_manual.rb
```diff
@@ -0,0 +1,56 @@
+require 'active_model'
+require 'valid_slug/pattern'
+
+class PublishingAPIRedirectedSectionsToParentManual
+  include ActiveModel::Validations
+  include Helpers::PublishingAPIHelpers
+
+  validates :manual_slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
```
davidbasalla: I suspect these validations are never called, as this class is only used to return the sections of a Manual and we don't call `valid?` anywhere

app/models/publishing_api_redirected_sections_to_parent_manual.rb
```diff
@@ -0,0 +1,56 @@
...
+    content_store: Services.content_store,
+    unless: -> { errors[:manual_slug].present? }
+
+  attr_accessor :manual_slug
+
+  def initialize(manual_slug)
+    @manual_slug = manual_slug
+  end
+
+  def sections
```
davidbasalla: As mentioned in the PR notes, this duplicates code somewhat, specifically the `sections` method from `PublishingApiRemovedManual`. Maybe we could move this method into the `PublishingApiManual` class, as that seems like a sensible place to have that method? Then we could get rid of this class entirely, and update `PublishingApiRemovedManual` to make use of it...? That's my theory at least :)

app/models/publishing_api_removed_manual.rb
sihugh: :+1:

app/models/publishing_api_removed_manual.rb
sihugh: Yep.  I flipped from originally using `start_index` because the query parameter Rummager uses is called `start`.  I agree that without that context it is more readable

lib/tasks/redirect_sections.rake
sihugh: 👍  (it's just so long! :-) )

app/models/publishing_api_redirected_sections_to_parent_manual.rb
sihugh: Yep.  I've removed this model now

app/models/publishing_api_redirected_sections_to_parent_manual.rb
sihugh: I've created a separate class `SectionRetriever` to remove the duplication a bit.  `PublishingAPIManual` seems a bit different from the other Manual models and the behaviour didn't seem to relate to it very well partly because we only want to create `Remove` or `Redirect` sections from the Rummager responses at present

sihugh:
I've rejigged it a bit in review of your comments @davidbasalla , what do you think?

lib/section_retriever.rb
```diff
@@ -0,0 +1,42 @@
...
+      new_query = rummager_section_query(start_index(search_response))
+
+      search_response = Services.rummager.search(new_query)
+      search_response["results"].map do |section_json|
+        sections << yield(section_json) if block_given?
+      end
+      return sections if all_sections_retrieved?(sections, search_response)
+    end
+  end
+
```
davidbasalla: From here on, all methods can be private I think, which would match the specs where we only test the `sections_from_rummager` method.

davidbasalla:
I like the new class, this approach seems good! However I'd prefer not to use the `yield` approach as I find it makes the code harder to decipher (possibly contentious personal opinion so happy to chat about it). Maybe instead of passing in a constructor (?) in the block, the `SectionRetriever` could become `SectionSlugRetriever` and just return an array of section slugs? So building the sections in the rake task would become:

```
    sections = SectionSlugRetriever.new(manual_slug).sections_from_rummager.map do |section_slug|
      PublishingAPIRedirectedSectionToParentManual.new(manual_slug, section_slug)
    end
```

Not 100% sure if that's better, but maybe easier to understand. Let's have a chat though!

lib/tasks/redirect_sections.rake
```diff
@@ -30,3 +13,54 @@ task :redirect_hmrc_section, [] => :environment do |_task, args|
...
+def redirect_and_output(item)
+  response = item.save!
+  if response.code == 200
+    puts "OK!"
+  else
+    output_error_message(response.raw_response_body)
+  end
+rescue => e
+  output_error_message(e.message)
+end
```
h-lame: It may be worth putting these two methods on an object as they're now defined globally whereas before they were defined only inside the rake task they were used by.  They've got fairly generic names and might bleed through into other parts of the app by mistake (or clobber some functions included by a dependency).

sihugh: Good point. It allows us to remove that same code from other rake tasks too.:+1:

lib/tasks/redirect_sections.rake
```diff
@@ -30,3 +13,37 @@ task :redirect_hmrc_section, [] => :environment do |_task, args|
...
+
+desc "Redirect all sections in a manual to the parent manual, takes manual slug"
+task :redirect_all_hmrc_sections_to_parent_manual, [] => :environment do |_task, args|
+  slugs = args.extras
+  if slugs.empty?
+    puts "Usage: rake redirect_all_hmrc_sections_to_parent_manual[manual-slug]"
+  else
+    manual_slug = slugs[0]
+
+    redirector = SectionRetriever.new(manual_slug)
```
davidbasalla: The variable name `redirector` here might be misleading. Maybe better to condense into one line? 

`sections = SectionRetriever.new(manual_slug).sections_from_rummager.map do |json|`

davidbasalla:
https://github.com/alphagov/hmrc-manuals-api/blob/redirect-all-sections/lib/tasks/redirect_sections.rake#L13 will need to be updated to use the new `TaskHelper`

davidbasalla:
All looks good, have fixed the last two comments while Simon's inoperational. 
Merging!

PR #164, opened 2017-01-27
Allow redirecting sections to other manuals

h-lame:
To support some of the zendesk tickets in: https://trello.com/c/2pxsDfWf/103-unpublish-hmrc-manual-sections

We already support redirecting a section to another section, or to it's parent manual.  This change allows us to redirect a section to another manual root.


davidbasalla:
Looks good to me! 👍

PR #169, opened 2017-04-03
Add documentation from the Ops Manual

klssmith:
We are moving documentation out of the Ops Manual - this commit adds the
documentation about hmrc-manuals-api from the Ops Manual to this repo.

[Trello card](https://trello.com/c/5x8NHLhf/40-merge-opsmanual-into-developer-docs)

### DIFF
--- a/docs/extended_documentation.md
+++ b/docs/extended_documentation.md
@@ -184,6 +184,17 @@ users to find.
 
 Use the [content-tagger](https://github.com/alphagov/content-tagger) app to add topic tags to manuals.
 
+## Removing published manuals
+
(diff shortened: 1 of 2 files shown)


docs/extended_documentation.md
```diff
@@ -60,6 +60,17 @@ The `<manual-slug>` and `<section_slug>` will be used as part of the GOV.UK URL
 
 [JSON Schema for sections](/public/section-schema.json)
 
+## Removing published manuals
```
jennyd: Most of the existing documentation here is intended for HMRC to explain to them how to publish manuals, and what's allowed in those requests, so I think this section should go much further down - probably just above `## Testing publishing in the GOV.UK development VM` which is also for GOV.UK developers, as this section is.

klssmith: Thanks - I've moved the new section to where you suggested.

PR #173, opened 2017-05-15
Always log request body

alexmuller:


tijmenb:
Maybe it's easier to always turn this logging on or off, no matter the environment? Would save another piece of configuration and way in which our apps differ from each other.

alexmuller:
The only difference is that it's disabled in the dev environment, and I don't know if that's intentional because it makes it too noisy to debug? Would need somebody who's worked on the app to comment I guess.

jennyd:
I think having it always turned on is likely to be fine. It's really useful to log the body so that we can more easily debug validation and formatting issues reported by HMRC - if we can see exactly what they're sending us it's much simpler to figure out where the problem lies. I think we were originally concerned that having the body in the logs might take up too much disk space, but the app gets minimal traffic in all environments other than Production and we haven't noticed it being a problem there.

alexmuller:
Less code works for me! Updated.

PR #174, opened 2017-05-17
Upgrade to Rails 5.1

leenagupte:
Trello cards: 
* https://trello.com/c/6EoCMZer
* https://trello.com/c/oW7yR5gL

## Motivation

hmrc-manuals-api is currently on version 4.2.7.1 of Rails, which is no longer supported. It needs to be upgraded to at least 5.0.2, preferably 5.1.0.

## Changes

The upgrade has been in two halves, first to Rails 5.0.2, and then a minor upgrade to 5.1 as it was believed that this would be easier.


leenagupte:
@thomasleese No I haven't. But good point! I'll see if I can deploy a branch today.

app/controllers/manuals_controller.rb
```diff
@@ -2,7 +2,8 @@ class ManualsController < ApplicationController
   before_action :parse_request_body, only: [:update]
 
   def update
-    manual = PublishingAPIManual.new(params[:id], @parsed_request_body)
+    permitted_params = params.permit!.to_h
```
thomasleese:
```ruby
permitted_params = params.permit(:id)
permitted_params[:id]
```

app/controllers/sections_controller.rb
```diff
@@ -2,7 +2,8 @@ class SectionsController < ApplicationController
   before_action :parse_request_body, only: [:update]
 
   def update
-    section = PublishingAPISection.new(params[:manual_id], params[:id], @parsed_request_body)
+    permitted_params = params.permit!.to_h
```
thomasleese:
```ruby
permitted_params = params.permit(:id, :manual_id)
permitted_params[:manual_id], permitted_params[:id]
```

PR #176, opened 2017-07-18
Use an environment variable for GDS-SSO config

leenagupte:
Trello card: https://trello.com/c/zpvWoNfW
Related to:
1. https://github.digital.cabinet-office.gov.uk/gds/deployment/pull/1372
2. https://github.digital.cabinet-office.gov.uk/gds/alphagov-deployment/pull/1335
3. https://github.com/alphagov/govuk-puppet/pull/6123

## Motivation

HMRC Manuals API uses a different way of configuring GDS SSO secrets to other GDS Rails apps, it should be updated to the follow the same conventions.


config/initializers/secret_token.rb
```diff
@@ -0,0 +1,12 @@
...
+# Your secret key for verifying the integrity of signed cookies.
+# If you change this key, all old signed cookies will become invalid!
+
+# Make sure the secret is at least 30 characters and all random,
+# no regular words or you'll be exposed to dictionary attacks.
+HMRCManualsAPI::Application.config.secret_token = 'ee51c777b16de4ba006fd4d5231deb3c941afe030c5a7374af5d0b2243d1fe0b06374361ad51ab19e4f3e422b1ab4f503aacc063ba2c078b2dee214b90ceb1b5'
+
+# Make sure your secret_key_base is kept private
+# if you're sharing your code publicly.
+HMRCManualsAPI::Application.config.secret_key_base = '4bfca259ad284cd807d5e463f1189df5a495698cc443ecfa0d8772e20d5cddf2b937ce35f6782ab27d02e81edb2ce3de79bbd6e10c1e4aba95d6d7c2ed5e8c91'
```
tijmenb: Since this application doesn't use cookies, you can replace it with `SecureRandom.hex` or something similar. You can disable session storage as well.

https://github.com/alphagov/collections/blob/master/config/initializers/session_store.rb

leenagupte: Ok. Thanks!

config/initializers/secret_token.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+# Your secret key for verifying the integrity of signed cookies.
+# If you change this key, all old signed cookies will become invalid!
+
+# Make sure the secret is at least 30 characters and all random,
+# no regular words or you'll be exposed to dictionary attacks.
+HMRCManualsAPI::Application.config.secret_token = 'ee51c777b16de4ba006fd4d5231deb3c941afe030c5a7374af5d0b2243d1fe0b06374361ad51ab19e4f3e422b1ab4f503aacc063ba2c078b2dee214b90ceb1b5'
```
tijmenb: I think Rails 4/5 doesn't use `config.secret_token`? But not sure. If it is used, it needs to be either random like `secret_key_base` or set using an environment variable.

leenagupte: Thanks. I'm going to take another look at the secret_token, secret_key_base config.

leenagupte:
@bevanloon True, but I decided to separate out the removal of values so that it was clear what was going and why.

PR #181, opened 2017-07-25
Send update type with put content

emmabeynon:
We are deprecating sending the `update_type` in publish requests made to the Publishing API, and now expect clients to send `update_type` in PUT requests instead.  We upgrade `gds-api-adapters` to bring this app up to date with the new expected behaviour.  Also, in this app we are currently passing 3 params to publish, so we must continue to send all 3 arguments until `update_type` is fully deprecated, hence why we change it to `nil` in the meantime.

https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk


emmabeynon:
> Looks good to me! Can I just confirm there was no instance of the republish update type?

Correct 💯

PR #183, opened 2017-08-08
Update InContentStoreValidator to validate on schema_name

emmabeynon:
I found that `InContentStoreValidator`was validating on `format`, however this has been deprecated in `govuk_content_schemas` (see https://github.com/alphagov/govuk-content-schemas/pull/620).  This meant that when attempting to redirect an HMRC manual section, the given section would be considered invalid as it has no format.  This updates `InContentStoreValidator` to validate against `schema_name` instead.  The formats initializer has also been removed as it has been replaced by a schema_name initializer.

TO DO:
- [x] Test redirecting an HMRC manual section on integration

https://trello.com/c/v2BJD3ER/1010-8-unpublish-sections-of-hmrc-manuals

### DIFF
--- a/spec/support/publishing_api_data_helpers.rb
+++ b/spec/support/publishing_api_data_helpers.rb
@@ -155,6 +155,7 @@ module PublishingApiDataHelpers
           'destination' => "/hmrc-internal-manuals/#{dest_manual_slug}/#{dest_section_slug}"
         }
       ],
+      'update_type' => 'major'
     }
   end
 
@@ -171,6 +172,7 @@ module PublishingApiDataHelpers
           'destination' => "/hmrc-internal-manuals/#{dest_manual_slug}"
         }
       ],
+      'update_type' => 'major'
     }
   end
 
@@ -187,6 +189,7 @@ module PublishingApiDataHelpers
           'destination' => "/hmrc-internal-manuals/#{manual_slug}"
         }
       ],
+      'update_type' => 'major'
     }
   end
 end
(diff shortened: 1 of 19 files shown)


spec/support/publishing_api_data_helpers.rb
```diff
@@ -155,6 +155,7 @@ def redirected_manual_section_to_other_manual_section_for_publishing_api(manual_
           'destination' => "/hmrc-internal-manuals/#{dest_manual_slug}/#{dest_section_slug}"
         }
       ],
+      'update_type' => 'major'
```
emmabeynon: You're right, this is a hangover from when we attempted to deprecate sending `update_type` with publish.

emmabeynon:
Thanks @kevindew, your comments should be addressed now.

kevindew:
Great thanks @emmabeynon, one thing I wonder now we've got the 2 SCHEMA and DOCUMENT constants, I think rummager should accept DOCUMENT fields rather than schema - not 100% sure on this though as I know very little rummager

emmabeynon:
> Great thanks @emmabeynon, one thing I wonder now we've got the 2 SCHEMA and DOCUMENT constants, I think rummager should accept DOCUMENT fields rather than schema - not 100% sure on this though as I know very little rummager

I think you're right, Rummager does reference document a lot, so I'll change that.

emmabeynon:
Tested redirecting an HMRC manual section on integration, which worked as expected 🚀

PR #247, opened 2018-02-23
Remove sidekiq and update dependencies

kevindew:
Should not be merged until https://github.com/alphagov/govuk-puppet/pull/7293 is deployed to production.

This removes sidekiq as it's no longer used and does all around maintenance to the app by fixing deprecation warnings, making Gemfile consistent and updating dependencies.

/cc @h-lame


kevindew:
@h-lame Yeah - I've not written it down somewhere. I'm not really sure where to put it. 

Currently it should be virtually impossible to trigger that error as Rails itself will raise it's own action dispatch param error (which converts to a 400) if we send malformed JSON in. So it looks like the controllers/routes need refactoring to a more modern rails param based approach and restrict non-json requests. This seems a somewhat tedious, low priority low value task so not sure where to even record it :-/

h-lame:
Fair enough, probably a more general case of an error that we might want to think about across all our apps.  I don't know where I'd write that down either.

PR #657, opened 2022-02-28
Upgrade rails version

BeckaL:
Bump major version of rails from 6.1 to 7.0


config/environments/production.rb
```diff
@@ -33,18 +33,18 @@
...
+  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
+  # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
   # Include generic and useful information about system operation, but avoid logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII).
-  config.log_level = :debug
+  config.log_level = :info
```
BeckaL, line 47: This was an automatic change from running rails app:update and I notice that we accepted it in the [content tagger PR](https://github.com/alphagov/content-tagger/pull/1272/files), but should we be changing this?

brucebolt: I think we should be keeping this at `:debug`, since we ship logs into Logit where we use them for debugging problems.

brucebolt (approved):
Looks good to me 👍 .

I've added a comment to your comment about the debug level.  I think it shouldn't be changed.

PR #702, opened 2022-07-07
Add parent manuals title to section

KludgeKML:
https://trello.com/c/NToovfjr/1349-add-parent-manuals-title-to-the-hmrc-section-content-item


app/models/publishing_api_manual.rb
```diff
@@ -87,6 +87,14 @@ def self.updates_path(manual_slug)
     "#{base_path(manual_slug)}/updates"
   end
 
+  def self.title(manual_slug)
+    content_id = GdsApi.publishing_api.lookup_content_id(base_path: base_path(manual_slug))
+    return "" unless content_id
```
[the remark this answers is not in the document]
KludgeKML: Have added this as a validation on the to_h method, so that it triggers a ValidationError when called in the controller.

spec/requests/manual_sections_spec.rb
```diff
@@ -38,6 +43,21 @@
...
+    stub_put_default_organisation(maximal_section_content_id)
+    stub_publishing_api_has_item(maximal_manual_for_publishing_api(content_id: maximal_manual_content_id, publication_state: "published", title: nil))
+
+    put maximal_section_endpoint,
+        params: maximal_section.to_json,
+        headers: { "CONTENT_TYPE" => "application/json",
+                   "HTTP_ACCEPT" => "text/plain",
+                   "HTTP_AUTHORIZATION" => "Bearer 12345" }
+    expect(response.status).to eq(422)
+  end
```
1pretz1, line 59: I don't think we're testing the intended behaviour, as the current scenario would be testing if a manual is returned from [Publishing API with a title of `nil`](https://github.com/alphagov/hmrc-manuals-api/pull/702/files#diff-cc160301fce2d1735f84ca14dbdc59e94531a796cc9d536cecc59522b030e048R94), after [successfully looking up the manual](https://github.com/alphagov/hmrc-manuals-api/pull/702/files#diff-cc160301fce2d1735f84ca14dbdc59e94531a796cc9d536cecc59522b030e048R91). Not sure if this is even possible as the `title` is a [required schema field](https://github.com/alphagov/govuk-content-schemas/blob/main/formats/shared/default_format.jsonnet#L7). 

Even though the validation does work as the test passes, it'd be better to test that the lookup returns nil, as this would be a realistic scenario. I think the below snippet would suffice:

```
  it "errors if the parent manual does not have a title" do
    stub_publishing_api_has_lookups({})

    put maximal_section_endpoint,
        params: maximal_section.to_json,
        headers: { "CONTENT_TYPE" => "application/json",
                   "HTTP_ACCEPT" => "text/plain",
                   "HTTP_AUTHORIZATION" => "Bearer 12345" }
    expect(response.status).to eq(422)
    expect(response.parsed_body["errors"]).to include("Missing manual related error message")
  end
```

app/models/publishing_api_section.rb
```diff
@@ -6,7 +6,14 @@ class PublishingAPISection
   include ActiveModel::Validations
   include Helpers::PublishingAPIHelpers
 
+  class ParentManualTitleValidator < ActiveModel::Validator
```
1pretz1, line 9: Could we add this to the validators directory? https://github.com/alphagov/hmrc-manuals-api/tree/main/app/validators

app/models/publishing_api_section.rb
```diff
@@ -6,7 +6,14 @@ class PublishingAPISection
...
+    def validate(record)
+      record.errors.add :manual_title, "Parent manual has no title" unless record.to_h.dig("details", "manual", "title")
```
1pretz1, line 11: We could probably add a clearer error message, maybe something like: `'Unable to find parent manual's title as manual doesn't exist in Publishing API for base_path: #{some_manual}'`

app/models/publishing_api_section.rb
```diff
@@ -6,7 +6,14 @@ class PublishingAPISection
...
+    end
+  end
+
   validates :to_h, no_dangerous_html_in_text_fields: true, if: -> { @section.valid? }
+  validates :to_h, parent_manual_title: true, if: -> { @section.valid? }
```
1pretz1, line 16: We probably want to add a unit test for this behaviour too: https://github.com/alphagov/hmrc-manuals-api/blob/597d3a6238560f5d98ba06798561eaa7017f325a/spec/models/publishing_api_section_spec.rb

1pretz1 (commented):
Thanks for adding the validation! Everything looks good (application code wise), just some small suggestions for test improvements / readability.

1pretz1 (commented):
Nice, thanks for working through the comments. Two minor comments and then we should be good to go 🚢

PR #707, opened 2022-07-21
Revert "Add parent manuals title to section"

1pretz1:
Reverts alphagov/hmrc-manuals-api#702

We're no longer want to add the manual title to the manual section. We originally wanted to add all the attributes a section needed when it renders, but this is more complicated than first thought.

The main reason we're reverting this change is because HMRC Manuals API could error when publishing a new manual, as the new code relies on a manual being published before a section and there is no validation ensuring this happens.

Dependant on: https://github.com/alphagov/govuk-content-schemas/pull/1108

Trello:
https://trello.com/c/ItYKGMqF/1464-revert-add-parent-manuals-title-to-the-hmrc-section-content-item

### DIFF
--- a/app/models/publishing_api_manual.rb
+++ b/app/models/publishing_api_manual.rb
@@ -87,14 +87,6 @@ class PublishingAPIManual
     "#{base_path(manual_slug)}/updates"
   end
 
-  def self.title(manual_slug)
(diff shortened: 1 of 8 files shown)


KludgeKML (approved):
LGTM

PR #730, opened 2022-10-03
Remove unnecessary `RUN mkdir /app` from Dockerfile.

sengi:
`WORKDIR` creates its directory anyway, and the mkdir was failing now that we're creating `/app` in govuk-ruby-base. Thought I'd fixed all of these before merging that change, but clearly I missed at least this one.


BeckaL (approved):
Thank you! ⭐

PR #872, opened 2023-11-02
Run tests on CI.

sengi:
This app uses rspec and not minitest, but we were only running the minitest action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,5 +21,4 @@ jobs:
 
   test-ruby:
     name: Test Ruby
-    uses: ./.github/workflows/minitest.yml
-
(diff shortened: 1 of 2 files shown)


sengi:
Oh, it does need the content schemas from publishing-api after all. My bad.

jkempster34:
Is this bit in GitHub's UI just leftover from the previous version? 
image

sengi:
That's the bit that needed changing — https://github.com/alphagov/govuk-saas-config/pull/176

jkempster34 (approved):

jkempster34:
> That's the bit that needed changing — [alphagov/govuk-saas-config#176](https://github.com/alphagov/govuk-saas-config/pull/176)

Ah, gotcha. Thank you!

PR #874, opened 2023-11-03
Make tests callable from publishing-api.

sengi:
Publishing API has [a pre-merge workflow][1] that calls these tests, so we need to be explicit with the checkout repo/ref.

[1]: https://github.com/alphagov/publishing-api/blob/3fdb7c2/.github/workflows/test-content-schemas-1.yml#L108-L113


jkempster34 (approved):

sengi:
Whoops, copy-pasto on the repo name there. Fixed.

PR #983, opened 2024-08-01
Allow previously redirected manuals to be removed

brucebolt:
If a manual or a section has been previously redirected, it has a schema of `redirect`. This means we cannot subsequently remove the manual or the redirect, as the new content item won't be valid.

Therefore allowing a schema of `redirect` for us to be able to remove a manual or a section that has previously been redirected.

[Trello card](https://trello.com/c/Shh2iPTU)

### DIFF
--- a/app/models/publishing_api_removed_manual.rb
+++ b/app/models/publishing_api_removed_manual.rb
@@ -6,7 +6,7 @@ class PublishingAPIRemovedManual
 
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
   validates_with InContentStoreValidator,
-                 schema_names: [MANUAL_SCHEMA_NAME],
+                 schema_names: [MANUAL_SCHEMA_NAME, "redirect"],
                  content_store: Services.content_store,
                  unless: -> { errors[:slug].present? }
 
--- a/app/models/publishing_api_removed_manual.rb
+++ b/app/models/publishing_api_removed_manual.rb
@@ -6,7 +6,7 @@ class PublishingAPIRemovedManual
 
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
   validates_with InContentStoreValidator,
-                 schema_name: MANUAL_SCHEMA_NAME,
+                 schema_names: [MANUAL_SCHEMA_NAME],
                  content_store: Services.content_store,
                  unless: -> { errors[:slug].present? }
 
--- a/app/validators/in_content_store_validator.rb
+++ b/app/validators/in_content_store_validator.rb
@@ -36,6 +36,6 @@ private
   end
 
   def wrong_schema_name_message(_record, content_item)
-    %{Exists in the content store, but is not a "#{schema_name}" schema (it's a "#{content_item['schema_name']}" schema)"}
+    %{Exists in the content store, but is not a "#{schema_name}" schema (it's a "#{content_item['schema_name']}" schema)}
   end
 end
--- a/app/validators/in_content_store_validator.rb
+++ b/app/validators/in_content_store_validator.rb
@@ -1,19 +1,19 @@
 class InContentStoreValidator < ActiveModel::Validator
-  attr_reader :schema_name, :content_store
+  attr_reader :schema_names, :content_store
 
   def initialize(options = {})
     super
-    raise "Must provide schema_name and content_store options to the validator" unless options[:schema_name] && options[:content_store]
-    raise 'Can\'t provide "gone" as a schema_name to the validator' if options[:schema_name] == "gone"
+    raise "Must provide schema_names and content_store options to the validator" unless options[:schema_names] && options[:content_store]
+    raise 'Can\'t provide "gone" as schema_names to the validator' if options[:schema_names].include?("gone")
 
-    @schema_name = options[:schema_name]
+    @schema_names = options[:schema_names]
     @content_store = options[:content_store]
   end
 
   def validate(record)
     content_item = fetch_content_item(record)
-    if content_item["schema_name"] != schema_name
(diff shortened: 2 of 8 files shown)


app/validators/in_content_store_validator.rb
```diff
@@ -1,19 +1,19 @@
...
-    @schema_name = options[:schema_name]
+    @schema_names = options[:schema_names]
     @content_store = options[:content_store]
   end
 
   def validate(record)
     content_item = fetch_content_item(record)
-    if content_item["schema_name"] != schema_name
-      record.errors.add(:base, wrong_schema_name_message(record, content_item))
+    unless schema_names.include?(content_item["schema_name"])
```
yndajas, line 15: Here you use `content_item["schema_name"]`

app/validators/in_content_store_validator.rb
```diff
@@ -35,7 +35,7 @@ def gone_message(_record, _content_item)
     'Exists in the content store, but is already "gone"'
   end
 
-  def wrong_schema_name_message(_record, content_item)
-    %{Exists in the content store, but is not a "#{schema_name}" schema (it's a "#{content_item['schema_name']}" schema)"}
+  def wrong_schema_names_message(_record, content_item)
+    %{Exists in the content store, but is not a "#{schema_names.join(',')}" schema (it's a "#{content_item['schema_names']}" schema)"}
```
yndajas, line 39: But here `content_item['schema_names']`. Is one of these incorrect? If it's the plural version, should it be joined here like `schema_names` is?

app/validators/in_content_store_validator.rb
```diff
@@ -1,19 +1,19 @@
 class InContentStoreValidator < ActiveModel::Validator
```
yndajas, line 1: Relatedly, should we have a spec for this class?

app/models/publishing_api_removed_manual.rb
```diff
@@ -6,7 +6,7 @@ class PublishingAPIRemovedManual
 
   validates :slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
   validates_with InContentStoreValidator,
-                 schema_names: [MANUAL_SCHEMA_NAME],
+                 schema_names: [MANUAL_SCHEMA_NAME, "redirect"],
```
yndajas, line 9: Took me a minute to understand the issue here, but is it that you need at least one schema name, and if we change it from the original name to redirect, then want to remove the redirect, it would have zero schema names instead of just the original one? So now we can simply remove the second/redirect schema name and the original schema name will remain?

spec/models/publishing_api_removed_manual_spec.rb
```diff
@@ -34,6 +34,12 @@
         expect(subject).to be_valid
       end
 
+      it 'is valid when the slugs represent a "redirect" piece of content' do
+        content_item = hmrc_manual_content_item_for_base_path(manual_path).merge("schema_name" => "redirect")
```
yndajas, line 38: Again does this need to be `schema_names` now, and the same in the `hmrc_manual_content_item_for_base_path` method and [elsewhere](https://github.com/search?q=repo:alphagov/hmrc-manuals-api%20schema_name&type=code)? Or does the code convert between the singular and multiple schema names somewhere?

yndajas (commented):
This might be fine, but I don't have a lot of context, so here are a few questions!

app/validators/in_content_store_validator.rb
brucebolt: This is the value of the content item we are checking, which will only have one value.  I've kept this singular for that reason (and as that's the name returned by content store).

app/validators/in_content_store_validator.rb
brucebolt: That should be singular.  I'll update it.

spec/models/publishing_api_removed_manual_spec.rb
brucebolt: This is singular, as it is effectively a stub for the content item returned to us by content store.

app/validators/in_content_store_validator.rb
brucebolt: I've added tests for this validator in https://github.com/alphagov/hmrc-manuals-api/pull/983/commits/f43a2a9f8d80ffc9c4952a3a82f1ddaa7a34a548.

app/models/publishing_api_removed_manual.rb
brucebolt: This validator is used in `PublishingAPIRemovedManual` and `PublishingAPIRemovedSection` to ensure the manual we're removing has the correct `schema_name` in Content Store.  However, if the manual has previously been redirected, the `schema_name` has been changed from `hmrc_manual` (or `hmrc_manual_section`) to `redirect`.  We therefore need to allow both names if we want to be able to remove previously redirected manuals.

app/validators/in_content_store_validator.rb
```diff
@@ -1,19 +1,19 @@
...
-  attr_reader :schema_name, :content_store
+  attr_reader :schema_names, :content_store
 
   def initialize(options = {})
     super
-    raise "Must provide schema_name and content_store options to the validator" unless options[:schema_name] && options[:content_store]
-    raise 'Can\'t provide "gone" as a schema_name to the validator' if options[:schema_name] == "gone"
+    raise "Must provide schema_names and content_store options to the validator" unless options[:schema_names] && options[:content_store]
+    raise 'Can\'t provide "gone" as a schema_names to the validator' if options[:schema_names].include?("gone")
```
yndajas, line 7: `as the schema_names` or `as schema_names`?

yndajas (approved):

app/validators/in_content_store_validator.rb
brucebolt: Oops.  I've fixed this.

PR #988, opened 2024-08-15
Follow upgrade guide for Rails 7.2

brucebolt:
This updates the application settings for Rails 7.2, which was upgraded in https://github.com/alphagov/hmrc-manuals-api/pull/985.

[Trello card](https://trello.com/c/KtQiL6h1)


public/404.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
```
yndajas, line 1: With Publishing API, I didn't include these. I assumed since it's an API, they were left out before. Would that be the case here too?

yndajas (approved):

public/404.html
brucebolt: That's a good point.  I've removed them from this commit and will not include them in the other API apps either.

PR #1028, opened 2024-11-22
Return 422 when request is unprocessable by Publishing API

brucebolt:
If the user provides data that is unprocessable by Publishing API (e.g. a base path with the incorrect content ID), we currently return a 500 error and log this to Sentry.

This is not optimal behaviour as there is no application fault here, so no action can be taken by developers after seeing this logged in Sentry.

Instead we should respond with a 422 and return the actual error to the user. Then they can correct the request they are making so it becomes valid.

[Trello card](https://trello.com/c/oAIfiY5s)

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -15,6 +15,8 @@ class SectionsController < ApplicationController
       end
     rescue ActionController::UnknownFormat
       render json: { status: "error", errors: "Invalid Accept header" }, status: :not_acceptable
+    rescue GdsApi::HTTPUnprocessableEntity => e
+      render json: { status: "error", errors: e }, status: :unprocessable_entity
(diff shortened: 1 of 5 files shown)


app/controllers/sections_controller.rb
```diff
@@ -13,8 +13,11 @@ def update
                  location: section.govuk_url
         end
       end
+      ### TODO: add this to the manuals controller too
```
callumknights, line 16: Looks like you've left this in, as I can see this has been added to the Manuals controller now.

callumknights (approved):
LGTM 👍 Other than the small comment I left regarding that #TODO

app/controllers/sections_controller.rb
brucebolt: Oops!

PR #1137, opened 2025-07-01
[PP-622] Allow top-level change notes on manuals

brucebolt:
We are allowing HMRC to include top-level change notes (i.e. notes for the manual itself, rather than a section).  This change allows HMRC to include these change notes by not including a Section ID in the change note.

### DIFF
--- a/spec/models/publishing_api_manual_spec.rb
+++ b/spec/models/publishing_api_manual_spec.rb
@@ -69,7 +69,6 @@ describe PublishingAPIManual do
     end
 
     context "valid_manual_without_change_note_titles" do
-      let(:attributes) { manual_without_change_note_titles }
       let(:path_1) { "/hmrc-internal-manuals/some-slug/abc567" }
       let(:path_2) { "/hmrc-internal-manuals/some-slug/abc555" }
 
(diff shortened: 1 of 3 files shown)


spec/models/publishing_api_manual_spec.rb
```diff
@@ -78,9 +77,22 @@
...
+          expect(subject.dig("details", "change_notes").second["title"]).to eq("Hmrc internal manuals some slug abc555")
+        end
+      end
+
+      context "when a section_id is not specified in a change note" do
+        let(:attributes) { manual_with_top_level_change_note }
+
+        it "adds the manual title to the title field of the change note without a section_id" do
+          expect(subject.dig("details", "change_notes").first["title"]).to eq("Employment Income Manual")
+          expect(subject.dig("details", "change_notes").second["title"]).to eq("Title of the previous Section that was changed")
```
yndajas, line 94: **question:** is there a reason for including and asserting on the section change notes when testing top-level change notes? To me it feels a bit out of place in this `valid_manual_without_change_note_titles` outer context, and in both this and the previous commit section change notes are covered in the sibling context

spec/support/test_data_helpers.rb
```diff
@@ -80,6 +80,33 @@ def maximal_manual(options = {})
...
+        child_section_groups: [],
+        change_notes: [
+          {
+            title: "Title of a Section that was changed",
+            section_id: nil,
+            change_note: "Description of changes",
+            published_at: "2014-01-23T00:00:00+01:00",
+          },
+          {
+            title: "Title of the previous Section that was changed",
```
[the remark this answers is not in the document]
brucebolt: Yes, I'll update the titles.

spec/models/publishing_api_manual_spec.rb
brucebolt: No particular reason.  I've removed them as they served no purpose.

spec/support/test_data_helpers.rb
```diff
@@ -80,6 +80,33 @@ def maximal_manual(options = {})
...
+      title: "Employment Income Manual",
+      content_id: maximal_manual_content_id,
+      description: "A manual about incoming employment",
+      public_updated_at: "2014-01-23T00:00:00+01:00",
+      update_type: "major",
+      details: {
+        child_section_groups: [],
+        change_notes: [
+          {
+            title: "Title of the section that was changed",
```
yndajas: **issue:** looks like this is still including a 'section' title on a top-level change note in the second commit

brucebolt: Good spot.  I've changed 'section' to 'manual' in this commit.

PR #1259, opened 2026-04-08
[PP-7424] Update Rails config for 8.1

yndajas:
This is a fairly straightforward update with decisions documented in commits. Notably the actual bump is absent since it was done in #1231, and there are no database schema changes since we don't have one


config/environments/production.rb
AgaDufrat: Not related to your changes but while you're there, would you mind correcting the healthcheck path on line 44:

```suggestion
config.silence_healthcheck_path = %w[/healthcheck/live, /healthcheck/ready]
```

AgaDufrat (approved):

config/environments/production.rb
yndajas: Is it preferable to use this silencing feature? We advised against this in the 8.0 guidance but I don't think we were particularly well-informed on that stance: https://docs.publishing.service.gov.uk/manual/guidance-for-upgrades-to-rails-8-0#configenvironmentproduction-rb

There's a mixture of not using this and using it with the incorrect or correct path(s) across alphagov repos: https://github.com/search?q=org%3Aalphagov+silence_healthcheck_path&type=code

AgaDufrat: Given how much we pay for Logit, and alternative ways to view healthcheck status I think it's sensible to set it. But it's probably worth asking Platform Engineering for opinion.

PR #1270, opened 2026-04-17
[PP-7633] Remove documentation pages

brucebolt:
This removes the documentation pages that were previously available at `/`, `/documentation` and `/readme`.

The reasons are:

- The documentation was broken and raising an error that was logged in Sentry. This has never been reported to us by HMRC, so we can assume they are not accessing this documentation through those paths.
- The documentation is available [through GitHub](https://github.com/alphagov/hmrc-manuals-api/blob/main/docs/extended_documentation.md) and [the developer docs](https://docs.publishing.service.gov.uk/repos/hmrc-manuals-api/extended_documentation.html), so there is no need to duplicate it again.
- The root page was showing `development` as the version number in all environments, which the team was not previously aware of, suggesting no-one is using this information which can alternatively be found in the Release app or Argo.


config/routes.rb
```diff
@@ -1,10 +1,6 @@
 Rails.application.routes.draw do
   # This strips out Rails' default .json/.xml format file extensions.
   scope format: false do
-    get "/", to: "root#index"
-    get "/documentation", to: "root#documentation"
-    get "/readme", to: redirect("/documentation")
-
```
AgaDufrat, line 7: **thought**: It would probably be nicer if those returned `410` (or redirected to other existing pages)
But given we have evidence that they are not used anyway, I think it fine.

AgaDufrat (approved):
Good to go!

config/routes.rb
brucebolt: Maybe these could return a redirect to either the dev docs or the page in GitHub 🤔

brucebolt: I'll remove them completely for now.  We can always put a redirect in place if anyone complains.

PR #1273, opened 2026-04-24
[PP-7620] Do not allow section breadcrumbs to self reference

brucebolt:
There have been cases where a section includes itself as a breadcrumb.

This is incorrect, as that breadcrumb should not be present.

Adding a validation to prevent this happening again in the future.

Only an error message, not code, has been added as this is consistent with the other validation errors. We may consider adding error codes in future work.

### DIFF
--- a/app/models/publishing_api_section.rb
+++ b/app/models/publishing_api_section.rb
@@ -8,6 +8,7 @@ class PublishingAPISection
   validates :manual_slug, :section_slug, format: { with: ValidSlug::PATTERN, message: "should match the pattern: #{ValidSlug::PATTERN}" }
   validate :incoming_section_is_valid
   validate :section_slug_matches_section_id, if: -> { @section.valid? }
+  validate :breadcrumbs_do_not_self_reference, if: -> { @section.valid? }
 
   attr_accessor :manual_slug, :section_slug, :section_attributes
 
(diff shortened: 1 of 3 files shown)


app/models/publishing_api_section.rb
```diff
@@ -128,4 +129,16 @@ def section_slug_matches_section_id
...
+  def breadcrumbs_do_not_self_reference
+    return unless section_attributes.dig("details", "breadcrumbs")
+
+    matching_breadcrumbs = section_attributes.dig("details", "breadcrumbs").map do |breadcrumb|
+      breadcrumb["section_id"] == section_attributes.dig("details", "section_id")
+    end
+
+    if matching_breadcrumbs.any?
+      errors.add(:base, "Breadcrumbs must not include the section they are added to")
+    end
```
AgaDufrat: This is spot on. We may want to consider a small refactor:
- call `.any?` with a block directly, as it will stop iterating as soon as it finds a match
- assign `details` to a variable, for readability

Sudo code (not tested):
```suggestion
    details = section_attributes["details"]
    return unless details&.dig("breadcrumbs")

    if details["breadcrumbs"].any? { |b| b["section_id"] == details["section_id"] }
      errors.add(:base, "Breadcrumbs must not include the section they are added to")
    end
```

AgaDufrat (approved):
LGTM. 
I'm assuming we're going to wait for an acknowledgment from the API users before deploying this breaking change.

brucebolt:
> I'm assuming we're going to wait for an acknowledgment from the API users before deploying this breaking change.

Yes, I'll inform HMRC of the change and also tell them which documents we've identified have this issue.

app/models/publishing_api_section.rb
brucebolt: Nice. I didn't realise that `any?` could take a block.

Rubocop has changed the formatting slightly, but it's basically the same as your suggestion now.

PR #1291, opened 2026-06-05
[DO NOT MERGE] Draft Assets endpoints spec for internal review

AgaDufrat:
The document outlines proposed endpoints for uploading, replacing, deleting, publishing, restoring and retrieving assets, along with several open design and implementation questions.

Please pay particular attention to the questions marked as REVIEW QUESTIONS throughout the document and add comments where you have a preference or concern.

**Note** that implementation of access limiting/auth bypass IDs may be out of scope for the MVP. As a potential interim workaround, HMRC users could retrieve assets from Signon and share them manually with reviewers who do not have Signon access.

### DIFF
--- a/docs/assets.md
+++ b/docs/assets.md
@@ -1,0 +1,328 @@
+# [DRAFT for internal review]
+
+> :warning: NOTE
+> Implementation of access limiting / auth bypass IDs maybe out of scope for the MVP
+
+> As a workaround HMRC could potentially login to Signon, download the file and email it to the person who needs to review it but doesn't have access to Signon.
+
+# Managing assets
+
+The assets endpoints allow clients to upload, replace, delete, publish, restore and retrieve assets associated with HMRC manuals.
+
+## Common headers
+
+### Request headers
+
+| Header          | Required | Description                                                  |
+| --------------- | -------- | ------------------------------------------------------------ |
+| `Authorization` | Yes      | Bearer token used to authenticate the request.               |
+| `Content-Type`  | No       | Defaults to `application/json`.                              |
+| `Accept`        | No       | Defaults to `application/json`.                              |
+
+### Response headers
+
+| Header         | Description                                                 |
+| -------------- | ----------------------------------------------------------- |
+| `Content-Type` | `application/json`                                          |
+
+---
+
+# Upload new asset
+
+Uploads a new asset after a clean virus scan result and creates a draft asset record.
+
+> REVIEW QUESTIONS
+> - Do we want to give the consumers the option to set asset to draft or do we enforce draft on first upload?
+> - Should we introduce 413 Payload Too Large response for uploads?
+> - Which subset of Asset Manager response fields (asset manager ID, file URL, state, replacement metadata, draft status, etc.) should be exposed through the HMRC Manuals API?
+
+## Request
+  
+```http
+POST /assets
+```
+
+### Form parameters
+
+| Parameter               | Required | Description                                                         |
+| ----------------------- | -------- | ------------------------------------------------------------------- |
+| `asset[file]`           | Yes      | File to upload.                                                     |
+| `asset[access_limited]` | TBC      | Whether access should be restricted. Defaults to HMRC organisation. |
+| `asset[draft]`          | No       | Whether the asset should remain in draft state. Defaults to `true`. |
+
+### Example request
+
+```bash
+curl -X POST \
+  http://hmrc-manuals-api.dev.gov.uk/assets \
+  -H "Authorization: Bearer <token>" \
+  -F "asset[file]=@logo.png"
+```
+
+### Success response
+
+```http
+201 Created
+```
+
+```json
+# See: https://github.com/alphagov/asset-manager/blob/main/app/presenters/asset_presenter.rb
+{
+  "id": "http://www.example.com/assets/6a216e0509c4d5e2e98bd731",
+  "name": "logo.png",
+  "content_type": "image/png",
+  "size": 82328,
+  "file_url": "https://assets.publishing.service.gov.uk/media/659bee95614fa20014f3a9f7/logo.png",
+  "state": "unscanned",
+  "draft": true,
+  "deleted": false,
+}
+```
+
+### Error responses
+
+| Status                     | Description                           |
+| -------------------------- | ------------------------------------- |
+| `401 Unauthorized`         | Authentication failed.                |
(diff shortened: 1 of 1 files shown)


docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+| `Content-Type` | `application/json`                                          |
+
+---
+
+# Upload new asset
+
+Uploads a new asset after a clean virus scan result and creates a draft asset record.
+
+> REVIEW QUESTIONS
+> - Do we want to give the consumers the option to set asset to draft or do we enforce draft on first upload?
```
brucebolt, line 35: Based on our discussions with HMRC's supplier, they'll only ever attach newly uploaded assets to draft documents (with the draft solely on their system, rather than draft on GOV.UK), so I think we should create all new assets as draft.  It'll make the documentation and implementation simpler for us.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+> - Should we introduce 413 Payload Too Large response for uploads?
```
brucebolt, line 36: Yes, but I'm not sure we have a defined size before the virus scanner fails 🤔

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+> - Which subset of Asset Manager response fields (asset manager ID, file URL, state, replacement metadata, draft status, etc.) should be exposed through the HMRC Manuals API?
```
brucebolt, line 37: We're talking to HMRC and their supplier on Wednesday, so we can find that out then.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Request
+  
+```http
+POST /assets
+```
+
+### Form parameters
```
brucebolt, line 45:
```suggestion
### Request parameters
```
Since the request will come from an API, instead of a form.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+| `401 Unauthorized`         | Authentication failed.                |
+| `413 Payload Too Large`    | Uploaded file exceeds permitted size. |
+| `422 Unprocessable Entity` | Asset could not be created.           |
+
+# Replace existing asset
+
+Replaces the file associated with an existing asset while retaining the asset identifier.
+
+> REVIEW QUESTIONS
+> - Is there a need to also redirect assets?
```
brucebolt, line 95: Possibly.  We can check with HMRC on Wednesday.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+| -------------------------- | ------------------------- |
+| `404 Not Found`            | Asset does not exist.     |
+| `422 Unprocessable Entity` | Asset replacement failed. |
+
+---
+
+# Delete asset
+
+> REVIEW QUESTIONS
+> - Should success response be 200 OK like in Asset Manager or 204 No Content?
```
brucebolt, line 147: I think 200 is correct, as the API is saying the request is ok.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+| Status          | Description                                   |
+| --------------- | --------------------------------------------- |
+| `404 Not Found` | Asset does not exist.                         |
+---
+
+# Publish draft asset
+
+Makes a draft asset publicly available.
+
+> Note: We don't have draft documents, they are immediately published https://github.com/alphagov/hmrc-manuals-api/blob/132edd90bce8061be40bef711ac40510a562f969/app/notifiers/publishing_api_notifier.rb#L6-L11 It may note be straight forward to automatically publish drafts assets associated with the draft document.
```
brucebolt, line 188: We don't need to deal with draft documents, as HMRC retain them solely in their system until they are published.  But the want to be able to embed draft assets from GOV.UK into their own system, to preview the asset in their drafts.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+> It will use Asset Managers Update endpoint
+
+> REVIEW QUESTIONS
+> - Should this be a separate endpoint or can we have Update asset endpoint that would allow users to update file and state based on params?
```
brucebolt, line 192: If we can just use the update endpoint, let's do that as it'll make the implementation easier for us.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+| `422 Unprocessable Entity`  | Asset couldn't be published. |
+
+---
+
+# Restore deleted asset
+
+Restores a previously deleted asset.
+
+> REVIEW QUESTIONS
+> - Should we add `409 Conflict`  | Asset is already published error response?
```
brucebolt, line 233: Sounds like a reasonable response to add.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
...
+> - Should restore be POST or PATCH?
```
brucebolt, line 234: I'd probably go for PATCH, as the endpoint is making a change to one parameter on the object.

docs/assets.md
```diff
@@ -0,0 +1,328 @@
+# [DRAFT for internal review]
+
+> :warning: NOTE
+> Implementation of access limiting / auth bypass IDs maybe out of scope for the MVP
+
+> As a workaround HMRC could potentially login to Signon, download the file and email it to the person who needs to review it but doesn't have access to Signon.
```
brucebolt, line 6: I don't think all the publishers of these manuals have Signon accounts, as they don't normally publish on GOV.UK (except for the HMRC manuals, through their Sharepoint).  So we'll need to consider how reviewers can access draft assets from the start.

brucebolt (commented):
This looks brilliant, great work 🎖️.  It's a nice concise solution to their problem.

I think I've added a comment (not necessarily an answer 😅) to each review question.  We're meeting HMRC on Wednesday, so can ask any outstanding questions then.

docs/assets.md
AgaDufrat: It feels like a separate work item for Asset Manager. I will write some thing.

docs/assets.md
AgaDufrat: I should've mentioned that it's POST in Asset Manager but it feels like it should be PATCH

brucebolt: Yes, it probably should be PATCH in Asset Manager too.  Possibly too late to change now, unless we go through and migrate all uses of POST to PATCH for this endpoint?

AgaDufrat:
Closing because it's been reviewed and we have a new version that was shared with HMRC for the review. 
Public documentation will be added to the repo when we add the new endpoints.

PR #1315, opened 2026-07-27
[PP-7792] Return 409 when Publishing API returns a conflict error

brucebolt:
If the user provides data that results in a conflict in Publishing API (e.g. they try to publish an already published edition), we currently return a 500 erorr and log this to Sentry.

This is not optimal behaviour as there is no application fault here, so no action can be taken by developers after seeing this logged in Sentry.

Instead we should respond with a 409 and return the actual error to the user. Then they can correct the request they are making so it becomes valid.
    
This mirrors the change made for 422 responses in #1028.

### DIFF
--- a/spec/support/publishing_api_helper.rb
+++ b/spec/support/publishing_api_helper.rb
@@ -1,5 +1,5 @@
 module PublishingApiHelper
-  def publishing_api_validation_error
-    stub_request(:any, /#{GdsApi::TestHelpers::PublishingApi::PUBLISHING_API_V2_ENDPOINT}\/.*/).to_return(status: 422)
+  def stub_publishing_api_returns_error(status)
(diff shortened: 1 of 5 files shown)


spec/support/publishing_api_helper.rb
```diff
@@ -1,5 +1,5 @@
 module PublishingApiHelper
-  def publishing_api_validation_error
-    stub_request(:any, /#{GdsApi::TestHelpers::PublishingApi::PUBLISHING_API_V2_ENDPOINT}\/.*/).to_return(status: 422)
+  def publishing_api_error(status)
```
yndajas: **suggestion:** in the tests it's not obvious to me that this is stubbing a request. Would it be worth adding a verb, e.g. `stub_publishing_api_error`?

yndajas (dismissed):

spec/support/publishing_api_helper.rb
brucebolt: Good idea.  I've renamed the method to `stub_publishing_api_returns_error`.

yndajas (approved):

PR #1316, opened 2026-07-27
[PP-7886] Add endpoint to allow upload of assets

brucebolt:
This implements the create asset endpoint detailed in [the API documentation](https://github.com/alphagov/hmrc-manuals-api/blob/main/docs/upload-and-manage-assets.md#upload-new-asset).

It allows both live and draft assets to be uploaded, then passed onto Asset Manager.

Requires https://github.com/alphagov/asset-manager/pull/2012.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,7 +7,6 @@ class ApplicationController < ActionController::Base
   skip_forgery_protection
 
   before_action :authenticate_user!
-  before_action :check_content_type_header
 
   rescue_from GdsApi::BaseError do |exception|
     GovukError.notify(exception)
@@ -30,7 +29,7 @@ private
     render json: { status: "error", errors: [message] }, status: :bad_request
   end
 
-  def check_content_type_header
+  def check_content_type_is_json
     if request.headers["Content-Type"] != "application/json"
       render json: { status: "error", errors: "Invalid Content-Type header" }, status: :unsupported_media_type
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,10 +13,10 @@ class ApplicationController < ActionController::Base
     if (exception.is_a?(GdsApi::HTTPErrorResponse) && (500..599).cover?(exception.code)) ||
         exception.is_a?(GdsApi::TimedOutException)
       message = "Service unavailable"
-      render json: { status: "error", errors: [message] }, status: :service_unavailable
+      error :service_unavailable, [message]
     else
       message = "Server error"
-      render json: { status: "error", errors: [message] }, status: :internal_server_error
+      error :internal_server_error, [message]
     end
   end
 
@@ -26,12 +26,16 @@ private
     @parsed_request_body = JSON.parse(request.body.read)
   rescue JSON::ParserError => e
     message = "Request JSON could not be parsed: #{e.message}"
-    render json: { status: "error", errors: [message] }, status: :bad_request
+    error :bad_request, [message]
   end
 
   def check_content_type_is_json
     if request.headers["Content-Type"] != "application/json"
-      render json: { status: "error", errors: "Invalid Content-Type header" }, status: :unsupported_media_type
+      error :unsupported_media_type, "Invalid Content-Type header"
     end
+  end
+
+  def error(status, errors)
+    render json: { status: "error", errors: }, status:
   end
 end
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -3,27 +3,38 @@ class AssetsController < ApplicationController
 
   def create
     asset = {
+      draft: asset_params[:draft].nil? || asset_params[:draft] != "false",
       file: asset_params[:file].tempfile,
     }
 
+    if asset[:draft]
+      asset[:auth_bypass_ids] = [SecureRandom.uuid]
+      asset[:auth_bypass_ids_expiry] = Time.zone.now + 30.days
+    end
+
     asset_manager_response = Services.asset_manager.create_asset(asset).to_h
+
+    output = {
+      _response_info: {
+        status: "ok",
+      },
+      asset_id: get_asset_id_from_url(asset_manager_response["file_url"]),
+      content_type: asset_manager_response["content_type"],
+      deleted: asset_manager_response["deleted"],
+      draft: asset_manager_response["draft"],
+      file_url: asset_manager_response["file_url"],
+      id: asset_manager_response["id"],
+      name: asset_manager_response["name"],
+      size: asset_manager_response["size"],
+      state: asset_manager_response["state"],
+    }
+
+    output[:file_url] = "#{asset_manager_response['file_url']}?token=#{asset[:auth_bypass_ids].first}" if asset[:auth_bypass_ids]
+    output[:preview_expiry] = asset[:auth_bypass_ids_expiry].iso8601 if asset[:auth_bypass_ids_expiry]
 
     respond_to do |format|
       format.json do
-        render status: :created, json: {
-          _response_info: {
-            status: "ok",
-          },
-          asset_id: get_asset_id_from_url(asset_manager_response["file_url"]),
-          content_type: asset_manager_response["content_type"],
-          deleted: asset_manager_response["deleted"],
-          draft: asset_manager_response["draft"],
-          file_url: asset_manager_response["file_url"],
-          id: asset_manager_response["id"],
-          name: asset_manager_response["name"],
-          size: asset_manager_response["size"],
-          state: asset_manager_response["state"],
-        }
+        render status: :created, json: output
       end
     end
   rescue ActionController::UnknownFormat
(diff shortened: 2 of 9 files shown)


app/lib/services.rb
```diff
@@ -17,7 +17,7 @@ def self.asset_manager
       @asset_manager ||= GdsApi::AssetManager.new(
         Plek.find("asset-manager"),
         bearer_token: ENV["ASSET_MANAGER_BEARER_TOKEN"] || "example",
-        timeout: 10,
+        timeout: 60,
```
callumknights, line 20: Good shout. It hadn't occurred to me to check Whitehall's timeout

app/controllers/application_controller.rb
```diff
@@ -27,12 +27,16 @@ def parse_request_body
...
   end
 
   def check_content_type_header
     if request.headers["Content-Type"] != "application/json" && !request.headers["Content-Type"].match?(/^multipart\/form-data/)
-      render json: { status: "error", errors: "Invalid Content-Type header" }, status: :unsupported_media_type
+      error :unsupported_media_type, "Invalid Content-Type header"
     end
   end
+
+  def error(status, errors)
```
callumknights, line 38: This is great!

callumknights (dismissed):
This all looks good to me 👍

app/lib/services.rb
yndajas, line 20: **question:** 

> Assets being sent to Asset Manager may take longer than 60 seconds to
upload, so increasing the timeout to 60 seconds.

Do you mean longer than 10 seconds?

app/controllers/application_controller.rb
```diff
@@ -31,7 +31,7 @@ def parse_request_body
   end
 
   def check_content_type_header
-    if request.headers["Content-Type"] != "application/json"
+    if request.headers["Content-Type"] != "application/json" && !request.headers["Content-Type"].match?(/^multipart\/form-data/)
```
yndajas: **question:** guessing there aren't any tests for the current behaviour to extend here?

app/controllers/assets_controller.rb
```diff
@@ -1,12 +1,18 @@
 class AssetsController < ApplicationController
   def create
     asset = {
+      draft: asset_params[:draft].nil? || asset_params[:draft],
       file: asset_params[:file].tempfile,
     }
 
+    if asset[:draft] == true
```
yndajas: **suggestion:** `== true` seems a little redundant

spec/requests/assets_spec.rb
```diff
@@ -2,14 +2,15 @@
...
+    let(:draft) { true }
     let(:asset_manager_response) do
       {
         _response_info: {
           status: "ok",
         },
         content_type: "text",
         deleted: "false",
-        draft: "false",
+        draft: draft,
```
yndajas: **question:** is `draft` intended to switch from a string to a Boolean?

spec/requests/assets_spec.rb
```diff
@@ -21,26 +22,84 @@
...
+      it "does not include an auth_bypass_id" do
+        post_multipart "/assets", {
+          asset: {
+            file: fixture_file_upload("asset.txt", "text/plain"),
+            draft: false,
+          },
+        }
+
+        parsed_response = JSON.parse(response.body).deep_symbolize_keys
+        expect(parsed_response).not_to include(have_key(:auth_bypass_id))
```
yndajas: **issue:** the response never has an `auth_bypass_id`. This should be testing for `file_url` and `preview_expiry`

Is the syntax correct? I wouldn't have thought the `include` should be there

spec/requests/assets_spec.rb
```diff
@@ -21,26 +22,84 @@
...
+        expect(parsed_response).to include(asset_id: "45678")
+      end
+
+      it "does not include an auth_bypass_id" do
+        post_multipart "/assets", {
+          asset: {
+            file: fixture_file_upload("asset.txt", "text/plain"),
+            draft: false,
+          },
+        }
```
yndajas: **suggestion:** this repeated code could maybe go in a `subject` for this context?

spec/requests/assets_spec.rb
```diff
@@ -21,26 +22,84 @@
...
+
+      it "responds with data from asset manager" do
+        post_multipart "/assets", {
+          asset: {
+            file: fixture_file_upload("asset.txt", "text/plain"),
+          },
+        }
+
+        parsed_response = JSON.parse(response.body).deep_symbolize_keys
+        expect(parsed_response).to include(asset_manager_response.except(:file_url))
```
yndajas: **question:** why the `except`? `file_url` seems to be included unconditionally

spec/requests/assets_spec.rb
```diff
@@ -21,26 +22,84 @@
...
+            file: fixture_file_upload("asset.txt", "text/plain"),
+          },
+        }
 
-      parsed_response = JSON.parse(response.body).deep_symbolize_keys
-      expect(parsed_response).to include(asset_manager_response)
-      expect(parsed_response).to include(asset_id: "45678")
+        parsed_response = JSON.parse(response.body).deep_symbolize_keys
+        expect(parsed_response).to include(file_url: "http://asset-manager.dev.gov.uk/media/45678/asset.txt?token=some-token")
+        expect(parsed_response).to include(preview_expiry: (Time.zone.now + 30.days).iso8601)
```
yndajas: **suggestion:** you could use `travel_to` to freeze time here and then assert on a known string value rather than repeating the logic

yndajas (commented):
Looks good - just a few things that might need a tidy up

app/lib/services.rb
brucebolt: Yes.  Have updated.

app/controllers/application_controller.rb
brucebolt: That's a good call.  The existing tests are on a controller-by-controller basis in the request specs.

Adding similar to the assets request spec made me realise I was missing `rescue ActionController::UnknownFormat` from the assets controller for checking the Accept header. I've added these tests and reworked the controller to raise and handle this error.

At the same time, I've also updated the Content Type check to be specific to the controller (i.e. JSON only for existing controllers, multipart only for the assets controller).

app/controllers/assets_controller.rb
brucebolt: This wasn't working because the value could've been either a string or boolean.

I've fixed this so the value is now always a boolean.

spec/requests/assets_spec.rb
brucebolt: Nope, I meant to add `to_s`.  Presumably why I wrote the variable name, rather than relying on hash value omission.

spec/requests/assets_spec.rb
brucebolt: Good spot.  I'd initially written all of this to return `auth_bypass_id` then re-read the documentation and realised it was appended to `file_url` 🤦.  Must've forgotten to update this test.

spec/requests/assets_spec.rb
brucebolt: It's because the value of `file_url` in the `let` block doesn't contain a token and we've got another test for that.

Do you think there's a better way I could communicate the reason for this?

spec/requests/assets_spec.rb
brucebolt: That's a really good idea.  I've changed this and also broken the expiry date into a separate spec (as we're testing different logic).

spec/requests/assets_spec.rb
yndajas: No I think this makes sense when read alongside `it "generates and includes a token in the file_url"`. Especially given the assertion here is on both key and value, and for `file_url` only the key will match

spec/requests/assets_spec.rb
brucebolt: This has made the tests a lot simpler to follow.

brucebolt: Alongside this, I've also introduced the `when Asset Manager responds with ok` context block in an earlier commit, which should make the subsequent commits easier to review as there is less diff.

spec/requests/assets_spec.rb
```diff
@@ -2,14 +2,68 @@
...
+        expect(parsed_response).to include(asset_manager_response)
+        expect(parsed_response).to include(asset_id: "45678")
+      end
+    end
+
+    it "errors if the Accept header is not application/json" do
+      post_multipart "/assets", {
+        asset: {
+          file: fixture_file_upload("asset.txt", "text/plain"),
+          draft: false,
```
yndajas: **todo:** I don't think this is meant to include the draft key - it predates the draft check

yndajas (commented):
Looks great - thanks for the changes. There's one quick thing to fix so no point approving yet, but I will do after!

spec/requests/assets_spec.rb
brucebolt: Ah, yes.  I've removed this.

yndajas (approved):

PR #1320, opened 2026-07-28
Add an Asset Manager service

callumknights:
Add a service method to allow connections to Asset Manager. This is currently gated behind a feature flag, with the intention that this feature only be used in integration for now. The plan is to remove that at a later date.

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -1,6 +1,7 @@
 require "gds_api/publishing_api"
 require "gds_api/search"
 require "gds_api/content_store"
+require "gds_api/asset_manager"
 
 module Services
   def self.publishing_api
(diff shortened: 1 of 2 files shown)


callumknights:
Depends on https://github.com/alphagov/govuk-helm-charts/pull/4371

app/lib/services.rb
```diff
@@ -1,6 +1,7 @@
 require "gds_api/publishing_api"
 require "gds_api/search"
 require "gds_api/content_store"
+require "gds_api/asset_manager"
```
brucebolt, line 4: Could these be put into alphabetical order? (I realise they aren't already, but will make them easier to work with)

app/lib/services.rb
```diff
@@ -11,6 +12,18 @@ def self.publishing_api
     )
   end
 
+  def self.asset_manager
+    if Rails.configuration.allow_asset_manager_requests
+      @asset_manager ||= GdsApi::AssetManager.new(
+        Plek.find("asset-manager"),
+        bearer_token: ENV["ASSET_MANAGER_BEARER_TOKEN"] || "example",
+        timeout: 10,
```
brucebolt, line 20: Note to myself, for a later PR: we'll have to assess whether 10 seconds will be long enough here.  I'll look into that in my PR to add the create asset endpoint.

brucebolt (approved):

app/lib/services.rb
callumknights: Will do in a separate PR shortly (as it will include the other `require` statements, and I don't want git to dismiss your review)

PR #1323, opened 2026-07-31
[PP-7887] Add endpoint to get Asset Metadata

callumknights:
This PR implements the GET asset endpoint detailed in [the API documentation](https://github.com/alphagov/hmrc-manuals-api/blob/main/docs/upload-and-manage-assets.md#upload-new-asset)

It allows HMRC Manuals API to retrieve Metadata about (currently) any asset in Asset Manager

Further revision to this may happen based on [this card](https://gov-uk.atlassian.net/jira/software/c/projects/PP/boards/398?selectedIssue=PP-7924)

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -1,5 +1,5 @@
 class AssetsController < ApplicationController
-  before_action :check_content_type_is_multipart
+  before_action :check_content_type_is_multipart, only: [:create]
 
   def create
     asset = {
@@ -45,6 +45,10 @@ class AssetsController < ApplicationController
     rescue GdsApi::HTTPUnprocessableEntity => e
       error :unprocessable_entity, e.message
     end
+  end
+
+  def show
+    render json: {}
   end
 
 private
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -48,7 +48,14 @@ class AssetsController < ApplicationController
   end
 
   def show
-    render json: {}
+    asset_manager_response = Services.asset_manager.asset(params[:id]).to_h
+    asset_manager_response.merge!({ asset_id: get_asset_id_from_url(asset_manager_response["file_url"]) })
+
+    render json: asset_manager_response
+  rescue GdsApi::HTTPNotFound
+    error :not_found, "Asset not found"
+  rescue GdsApi::HTTPForbidden
+    error :forbidden, "Access to asset is forbidden"
   end
 
 private
--- a/spec/requests/assets_spec.rb
+++ b/spec/requests/assets_spec.rb
@@ -3,6 +3,18 @@ require "rails_helper"
 describe "assets resource" do
   include ActiveSupport::Testing::TimeHelpers
 
+  describe "GET /assets/:id" do
+    subject do
+      get "/assets/123456"
+    end
+
+    it "responds with 200" do
+      subject
+
+      expect(response.status).to eq(200)
+    end
+  end
+
   describe "POST /assets" do
     let(:draft) { true }
     let(:asset_manager_response) do
--- a/spec/requests/assets_spec.rb
+++ b/spec/requests/assets_spec.rb
@@ -4,14 +4,71 @@ describe "assets resource" do
   include ActiveSupport::Testing::TimeHelpers
 
   describe "GET /assets/:id" do
-    subject do
-      get "/assets/123456"
+    let(:asset_manager_response) do
+      {
+        _response_info: {
+          status: "ok",
+        },
+        content_type: "text",
+        deleted: "false",
+        draft: "false",
+        file_url: "http://asset-manager.dev.gov.uk/media/123456789/asset.txt",
+        id: "http://asset-manager/assets/123456789",
+        name: "asset.txt",
+        size: "12",
+        state: "clean",
+      }
     end
 
-    it "responds with 200" do
-      subject
+    subject do
+      get "/assets/123456789"
+    end
 
-      expect(response.status).to eq(200)
+    context "when asset manager responds with ok" do
+      before do
+        allow(Services.asset_manager).to receive(:asset).and_return(asset_manager_response.deep_stringify_keys)
+      end
+
+      it "responds with 200" do
+        subject
+
+        expect(response.status).to eq(200)
+      end
+
+      it "responds with data from asset manager" do
+        subject
+
+        parsed_response = JSON.parse(response.body).deep_symbolize_keys
+
+        expect(parsed_response).to include(asset_manager_response)
+        expect(parsed_response).to include(asset_id: "123456789")
+      end
+    end
+
+    context "when asset manager responds with GdsApi::HTTPNotFound" do
+      before do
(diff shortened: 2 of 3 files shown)


spec/requests/assets_spec.rb
```diff
@@ -4,14 +4,58 @@
...
+        expect(response.status).to eq(200)
+      end
+
+      it "responds with data from asset manager" do
+        subject
+
+        parsed_response = JSON.parse(response.body).deep_symbolize_keys
+
+        expect(parsed_response).to include(asset_manager_response)
+        expect(parsed_response).to include(asset_id: "123456789")
```
yndajas, line 44: **question:** in the `POST /assets` tests, what's the ID in the Asset Manager response and why is it different to the one from the file URL?

Why is the ID in these new tests shaped differently (like a URL, but with the same ID as in the file URL)?

spec/requests/assets_spec.rb
```diff
@@ -4,14 +4,58 @@
...
+      end
+    end
+
+    context "when asset manager responds with GdsApi::HTTPNotFound" do
```
yndajas, line 48: **issue:** looks like there's no test for the forbidden response

app/controllers/assets_controller.rb
```diff
@@ -48,7 +48,14 @@ def create
   end
 
   def show
-    render json: {}
+    asset_manager_response = Services.asset_manager.asset(params[:id]).to_h
+    asset_manager_response.merge!({ asset_id: get_asset_id_from_url(asset_manager_response["file_url"]) })
```
brucebolt: Would it be better to explicitly list the fields returned here?  Otherwise changes in Asset Manager could affect the output of this endpoint.

callumknights: Okay that's fine. I wasn't sure about explicitly listing the fields. And mainly shy'ed away cause I wasn't making many changes to the data, and the current response from asset-manager matched. But you're right, future changes could alter that.

spec/requests/assets_spec.rb
callumknights: Will add 👍

spec/requests/assets_spec.rb
callumknights: You'd have to check with @brucebolt about the intention of the differing ID's in the `POST /assets` tests, but I took it as a demonstration that `id` and `asset_id` don't necessarily have to be the same (though is practise I they would be). The reason I did it this way was cause that's exactly what a response looks like from asset manager, so I was just keeping it as close to reality as possible.

brucebolt: My understanding is that Asset Manager returns a `id` field which is actually a URL.  But my tests are agnostic to the value that is returned, as HMRC Manuals API doesn't really care what the value is, just that it needs to be passed onto the client.

I chose a different value to the number in the URL so it would be clear that they aren't the same ID.

callumknights: "My understanding is that Asset Manager returns a id field which is actually a URL." This is exactly the case. I stuck with the URL in the test for a similar reason, but with the bonus of being canonical to the asset-manager response.

app/controllers/assets_controller.rb
```diff
@@ -48,7 +48,28 @@ def create
...
+      },
+      asset_id: get_asset_id_from_url(asset_manager_response["file_url"]),
+      content_type: asset_manager_response["content_type"],
+      deleted: asset_manager_response["deleted"],
+      draft: asset_manager_response["draft"],
+      file_url: asset_manager_response["file_url"],
+      id: asset_manager_response["id"],
+      name: asset_manager_response["name"],
+      size: asset_manager_response["size"],
+      state: asset_manager_response["state"],
```
brucebolt: Since these are the same as the other endpoint, could they be moved into somewhere shared?  Maybe a presenter (like it's done in Asset Manager itself)?

app/controllers/assets_controller.rb
```diff
@@ -1,5 +1,5 @@
 class AssetsController < ApplicationController
-  before_action :check_content_type_is_multipart
+  before_action :check_content_type_is_multipart, only: [:create]
```
brucebolt, line 2: Do we need `check_content_type_is_json` for the `show` method (as is the case for the other controllers)?

spec/requests/assets_spec.rb
```diff
@@ -3,6 +3,75 @@
 describe "assets resource" do
   include ActiveSupport::Testing::TimeHelpers
 
+  describe "GET /assets/:id" do
```
brucebolt, line 6: Is a similar test needed for `errors if the Accept header is not application/json` and `errors if the Content-Type header is not application/json` as we have with the other endpoint and controllers in this app?

app/controllers/assets_controller.rb
callumknights: A [content type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) shouldn't even be present on a GET request. I can't find an explicit reference that this isn't allowed, but given the request in this case lacks a message body, I'd think [RFC3875 Section 4.1.3](https://www.ietf.org/rfc/rfc3875#section-4.1.3) applies here

spec/requests/assets_spec.rb
callumknights: These tests shouldn't be needed for the same reasons as https://github.com/alphagov/hmrc-manuals-api/pull/1323#discussion_r3704850886

app/controllers/assets_controller.rb
callumknights: Good shout! I'll do that

app/controllers/assets_controller.rb
brucebolt: Makes sense.  Looking at the existing tests, we only use them for PUT requests.

spec/requests/assets_spec.rb
brucebolt: Cool, that explains why the existing tests are only present for the PUT requests.

app/controllers/assets_controller.rb
callumknights: Separate discussions on slack have lead to the conclusion it's better to delegate the contents of this response to Asset Manager, as being specific like the above code opens the door to breaking changes in the future. Changes to reflect this will follow.

app/controllers/assets_controller.rb
brucebolt: @callumknights and me have chatted on Slack and made a decision to delegate this to Asset Manager (for both this endpoint and the one that has already been implemented).  The fields are the same (except the `asset_id`), so we can have a shared method to override that.

callumknights: https://github.com/alphagov/hmrc-manuals-api/pull/1323#discussion_r3705487117

callumknights: Changes pushed. A further refactor for the `#create` endpoint, and abstracting out any shared logic will follow, but outside the scope of this PR.

brucebolt: I'm happy to refactor the `create` method, as I originally implemented it.

brucebolt (approved):

PR #1325, opened 2026-08-03
Delegate response to Asset Manager

brucebolt:
In https://github.com/alphagov/hmrc-manuals-api/pull/1323, we made the decision to delegate the asset management response fields to Asset Manager, to avoid duplication of code between the two apps.

This does the same for the create asset endpoint, and moves the code into a shared helper.


callumknights (approved):
Fantastic! 👍

PR #1330, opened 2026-08-11
[PP-7889] Add endpoint to regenerate an asset's access token

brucebolt:
This will allow HMRC to request a new access token for draft assets, when their existing one expires or they need to expire the old token.

Implemented in accordance with the [draft API documentation](https://github.com/alphagov/hmrc-manuals-api/blob/main/docs/upload-and-manage-assets.md#regenerate-draft-asset-access).


yndajas (dismissed):
A couple of non-blocking design suggestions

yndajas (approved):

PR #1331, opened 2026-08-12
[PP-7884] Add asset update endpoint

callumknights:


app/controllers/assets_controller.rb
```diff
@@ -32,10 +32,37 @@ def create
   end
 
   def update
-    respond_to do |format|
-      format.json do
-        render json: {}
+    asset = {
+      draft: cast_boolean(asset_params[:draft]),
```
brucebolt: What happens if a value for `draft` isn't supplied by the client?  None of the tests cover this scenario.

app/controllers/assets_controller.rb
```diff
@@ -68,10 +95,15 @@ def check_content_type_is_multipart
     end
   end
 
+  def cast_boolean(value)
+    ActiveModel::Type::Boolean.new.cast(value)
+  end
```
brucebolt: Does this need to be a method, it's only being used once?

spec/requests/assets_spec.rb
```diff
@@ -220,15 +220,172 @@
...
+    end
+
     subject do
-      put_multipart "/assets/123456", {}
+      put_multipart "/assets/123456789", {
+        asset: {
+          file: fixture_file_upload("asset.txt", "text/plain"),
+        },
+      }
+    end
```
brucebolt: This is not being used anywhere, as each sub-context sets it's own subject.

spec/requests/assets_spec.rb
```diff
@@ -220,15 +220,172 @@
...
+
+    context "when asset manager responds with ok" do
+      before do
+        allow(Services.asset_manager).to receive(:update_asset).and_return(asset_manager_response.deep_stringify_keys)
+      end
+
+      context "when the asset is not draft" do
+        let(:draft) { false }
```
brucebolt: The memoization helpers here and on line 335 are being used in `asset_manager_response` but not in the subject (the value is hardcoded again). I'd suggest using the value for `draft` in the subject too.

app/controllers/assets_controller.rb
```diff
@@ -32,10 +32,37 @@ def create
...
+      error :not_acceptable, "Invalid Accept header"
+    rescue GdsApi::HTTPPayloadTooLarge
+      error :content_too_large, "Content exceeds maximum permitted size"
+    rescue GdsApi::HTTPUnprocessableEntity => e
+      error :unprocessable_entity, e.message
+    rescue GdsApi::HTTPNotFound
+      error :not_found, "Asset not found"
+    rescue GdsApi::HTTPForbidden
+      error :forbidden, "Access to asset is forbidden"
     end
```
brucebolt: There's no tests for the exceptions on lines 58 to 66.

brucebolt: You've added in some additional exceptions here which aren't listed in the documentation (which is good, and we should keep them as they are definitely possible scenarios).  Can you also update the documentation to reflect these possible responses.

spec/requests/assets_spec.rb
```diff
@@ -220,15 +220,172 @@
...
+      put_multipart "/assets/123456789", {
+        asset: {
+          file: fixture_file_upload("asset.txt", "text/plain"),
+        },
+      }
+    end
+
+    context "when asset manager responds with ok" do
+      before do
+        allow(Services.asset_manager).to receive(:update_asset).and_return(asset_manager_response.deep_stringify_keys)
```
brucebolt: I think it would be beneficial to test what we're actually sending to Asset Manager, rather than stubbing the method call for any arguments, as this stub meant the `ArgumentError` never got noticed.

I see none of the other endpoints test for this either, which is an oversight.  I think that we should implement this change in this PR for this endpoint (as there's multiple combinations of values we could send) and I'll write a ticket to add this to the existing endpoints.

There are test helpers in gds-api-adapters that should make this testing easy to implement.

spec/support/json_request_helpers.rb
```diff
@@ -12,6 +12,7 @@ def put_multipart(path, attrs, headers = {})
     default_headers = {
       "HTTP_ACCEPT" => "application/json",
       "HTTP_AUTHORIZATION" => "Bearer 12345678",
+      "CONTENT_TYPE" => "multipart/form-data",
```
brucebolt: Could this be merged with the commit where the `put_multipath` method is first introduced?

brucebolt (commented):
The logic mostly looks ok to me, but my requests to the endpoint fail with a 500 response and the following error is logged:

```rb
ArgumentError (wrong number of arguments (given 1, expected 2)):
 
app/controllers/assets_controller.rb:44:in 'AssetsController#update'
```

app/controllers/assets_controller.rb
callumknights: Yeah this was on my list to do. Given one (forbidden) was introduced with the Asset Manager work, and other (PayloadTooLarge)seems like a natural consequence of this endpoint, given it also exists for the `create` endpoint.

app/controllers/assets_controller.rb
callumknights: _Maybe not_, but I wasn't a fan of the syntax of having a direct call to an ActiveModel type in line where the params are modified. So I thought it looked cleaner as a method, happy to be overruled though.

spec/support/json_request_helpers.rb
callumknights: Probably, I left it as it's own commit to document why I'd added it, as it took me a while to investigate and I figured it would be better to have that information somewhere obvious. As an aside, the `as: :multipart` token can actually be removed, it's not doing anything as `:multipart` is not a registered test encoder. I can do that.

app/controllers/assets_controller.rb
callumknights: Good point about the tests, I'll add one in. But for explanation, the point of the `cast_boolean` method is it preserves `nil`'s, so if not provided, it gets filtered out of the `asset_params` and no changes to draft state are made.

brucebolt: Thanks for the explanation.  Adding a test will definitely be useful, to prevent any regression in the future if the controller code gets refactored.

app/controllers/assets_controller.rb
callumknights: Addressed [here](https://github.com/alphagov/hmrc-manuals-api/pull/1336)

spec/support/json_request_helpers.rb
callumknights: Having thought about this more, I'll merge it with the first commit. But include the commit message with my investigation in that commit, as I still think capturing that is important.

spec/requests/assets_spec.rb
callumknights: I've addressed this and used it in lower down error cases. As well as the 'no draft' test.

app/controllers/assets_controller.rb
callumknights: A 'no draft' test has now been added.

spec/requests/assets_spec.rb
callumknights: This has been addressed. Thanks for pointing it out.

app/controllers/assets_controller.rb
callumknights: These now have tests. Also using the base subject.

spec/requests/assets_spec.rb
callumknights: This is the main final thing for me to do (just commenting this for visibility). I've addressed on the other above testing missteps, now I've got them straight, I'll go through and refactor to use the TestHelpers provided by gds-api-adapters

spec/support/json_request_helpers.rb
callumknights: This is now addressed as well. I've also removed the redundant `:multipart` token.

spec/requests/assets_spec.rb
callumknights: I've refactored these tests to use the Asset Manager helpers. I've also added some additional test helpers in [this PR](https://github.com/alphagov/gds-api-adapters/pull/1415), which is why the final commit currently uses a remote git branch for the `gds-api-adapters` gem. I'll remove that once the new helpers are released.

spec/requests/assets_spec.rb
```diff
@@ -221,15 +221,262 @@
...
+          expect(decoded_token_payload_from_url(parsed_response[:file_url])).to eq(expected_decoded_token)
+        end
+
+        it "includes a preview expiry date 30 days in the future" do
+          travel_to Time.zone.local(2026, 1, 1, 0, 0, 1)
+          subject
+
+          parsed_response = JSON.parse(response.body).deep_symbolize_keys
+          expect(parsed_response).to include(preview_expiry: Time.zone.local(2026, 1, 31, 0, 0, 1).iso8601)
+        end
```
brucebolt: Suggestion: these same tests are now repeated three times within this spec file, maybe they could exist as a shared example so we don't need to repeat the same code?  This could be a separate commit in this PR.

spec/requests/assets_spec.rb
```diff
@@ -221,15 +221,262 @@
...
+
+      context "when request params includes a replacement_id" do
+        let(:draft) { false }
+        let(:additional_attributes) { { replacement_id: "987654321" } }
+
+        subject do
+          put_multipart "/assets/123456789", {
+            asset: {
+              replacement_id: "987654321",
+              draft: draft,
```
brucebolt: As it is an optional value, I'm not sure setting a value for `draft` is providing any benefit here, as there are already tests for the draft value elsewhere.  Then you can remove the "when a draft value is not given by the client" test, because the situation of no draft value being supplied is already being tested here.

spec/requests/assets_spec.rb
```diff
@@ -221,6 +223,265 @@
...
+        expect(response.body).to include("Asset does not exist")
+      end
+    end
+
+    context "when Asset Manager responds with GdsApi::HTTPForbidden" do
+      before do
+        stub_asset_manager_responds_forbidden("123456789")
+      end
+
+      it "returns a 413 response" do
```
brucebolt:
```suggestion
      it "returns a 403 response" do
```

spec/requests/assets_spec.rb
```diff
@@ -221,15 +221,262 @@
...
+        end
+
+        it "includes a preview expiry date 30 days in the future" do
+          travel_to Time.zone.local(2026, 1, 1, 0, 0, 1)
+          subject
+
+          parsed_response = JSON.parse(response.body).deep_symbolize_keys
+          expect(parsed_response).to include(preview_expiry: Time.zone.local(2026, 1, 31, 0, 0, 1).iso8601)
+        end
+      end
```
brucebolt: The draft tests could be moved into a single context, so you only need to declare the subject once.  I'd also not include the value for `file` in the request, since these tests are specifically looking at the value for `draft`.

```rb
context "when the client provides a value for draft" do
  subject do
    put_multipart "/assets/123456789", {
      asset: {
        draft: draft,
      },
    }
  end

  context "when the asset is not draft" do
    let(:draft) { false }

    [specs here]
  end

  context "when the asset is draft" do
    before do
      allow(SecureRandom).to receive(:uuid).and_return("some-token")
    end

    [specs here]
  end
end
```

brucebolt (commented):
Thanks for making those changes.

The application code is looking good to me.  I've got a few suggestions about the tests.

I think it'd be better the break the tests down into a few categories:
- when the client supplies a value for draft
- when the client supplies a new file
- when the client supplies a replacement_id
- all the error cases

Doing this will make the tests easier to follow and explicitly define what is being tested each time (e.g. at the moment there is no explicit test for when the client supplies a new file, as it is named "when a draft value is not given by the client").

spec/requests/assets_spec.rb
brucebolt: Thanks.  This method of stubbing the requests is definitely more robust than what we were doing previously.

spec/requests/assets_spec.rb
callumknights: Thanks for this. I've restructured everything in more or less the way you have suggested. I've moved the 'new file' context to the very top of the specs, below the stubbed response, as it's the only one using the top level subject, and I feel that's the easiest to read. That way every other context below defines it's own subject request at some point. At least until you get to the error cases that use the top level one.

spec/requests/assets_spec.rb
callumknights: Working on this extra commit now

spec/requests/assets_spec.rb
yndajas: > I think it would be beneficial to test what we're actually sending to Asset Manager

Is this bit done? I'm thinking about this for the restore endpoint. It seems useful but I'm wondering whether it belongs in a request spec. Can't tell where it's done here from a quick glance

yndajas: Is it by using the stubs with a specific asset ID?

brucebolt: I hadn't realised these helpers weren't stubbing specific payload requests (only the ID and response).  That's probably something we should do.

callumknights:
All testing comments should now be addressed. I've added an extra commit with a test that checks if the request params originally passed in make it through to Asset Manager as intended, as discussed [here](https://gds.slack.com/archives/C04VBRRSCJW/p1787140664688759)

spec/requests/assets_spec.rb
```diff
@@ -5,6 +5,24 @@
   include ActiveSupport::Testing::TimeHelpers
   include GdsApi::TestHelpers::AssetManager
 
+  shared_examples "includes a draft response token" do
```
brucebolt: I like this shared helper.  It'll be really useful if we ever need to vary how drafts work, as fewer tests will need updating.

It's probably worth updating the regenerate access tests to use this shared example too.

spec/requests/assets_spec.rb
```diff
@@ -1,10 +1,21 @@
...
+  include GdsApi::TestHelpers::AssetManager
 
   let(:parsed_response) { JSON.parse(response.body).deep_symbolize_keys }
 
+  shared_examples "passes request params to Asset Manager" do
+    it "preserves the request params" do
+      expected_params = request_params.fetch(:asset).dup
+      expected_params[:file] = anything if expected_params.key?(:file)
+
+      expect(Services.asset_manager).to have_received(:update_asset).with("123456789", hash_including(expected_params))
```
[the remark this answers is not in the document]
callumknights: I've thought about it a fair bit and I'm not so sure that's the correct approach for this specific shared example. The whole point of this example is just to test the controller passes the `request_params` through as expected, so in my head, I think it's fine for the limit of it's scope to be the call to `Services.asset_manager` as that accomplished that task just fine. One thing I do think though, is currently the `create` endpoint lacks the same test, and given that's the only other endpoint that relies heavily on request_params being passed through, I think that's valuable. 

On my branch, freshly rebased, I can currently comment out https://github.com/alphagov/hmrc-manuals-api/blob/main/app/controllers/assets_controller.rb#L8 and I get no (additional) test failures. 

So I'd instead propose either, memoizing the asset manager method being used in each test endpoint test block, so it can be used in this shared example. Or using the shared_example with an argument, which would be that method, in case we are collectively sick of memoization helpers.

Note: In theory the we could still use the stub helpers for that, by writing a test where they expect a request with a given method and body based on the request params, however I can't see an easy way to make it flexible enough for a shared example, so would result in a lot of code duplication, and like I said above, I question the value of that anyway.

spec/requests/assets_spec.rb
```diff
@@ -229,15 +240,228 @@
     end
   end
 
-  describe "PUT /asset/:id" do
+  describe "PUT /assets/:id" do
```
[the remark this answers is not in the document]
callumknights: This is now split up appropriately, 14 commits in all to make each substantive change it's own commit

spec/requests/assets_spec.rb
```diff
@@ -229,15 +240,228 @@
...
+
+        it "responds with data from Asset Manager" do
+          expect(parsed_response).to include(asset_id: "123456789")
+        end
+
+        it_behaves_like "passes request params to Asset Manager"
+      end
+
+      context "when the client provides a value for draft" do
+        let(:request_params) { { asset: { draft: draft, file: fixture_file_upload("asset.txt", "text/plain") } } }
```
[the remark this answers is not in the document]
callumknights: Changed 👍

spec/requests/assets_spec.rb
```diff
@@ -229,15 +240,228 @@
...
+          expect(response).to have_http_status(:ok)
+        end
+
+        it "responds with data from Asset Manager" do
+          expect(parsed_response).to include(asset_id: "123456789")
+        end
+
+        it "includes the replacement_id in the response" do
+          expect(parsed_response).to include(replacement_id: "987654321")
+        end
```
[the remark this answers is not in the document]
callumknights: Removed 👍

spec/requests/assets_spec.rb
```diff
@@ -229,15 +240,228 @@
...
+    end
+
+    context "when Asset Manager responds with GdsApi::HTTPUnprocessableEntity" do
+      before do
+        stub_asset_manager_responds_unprocessable
+
+        subject
+      end
+
+      it "returns with 422 Unprocessable Entity" do
```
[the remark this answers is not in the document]
callumknights: This is now resolved as well 👍

spec/requests/assets_spec.rb
```diff
@@ -462,6 +462,22 @@
       end
     end
 
+    context "when the Accept header is not application/json" do
+      subject do
+        put_multipart "/assets/12345", request_params, { "HTTP_ACCEPT" => "text/plain" }
+      end
+
+      before do
+        allow(Services.asset_manager).to receive(:update_asset).and_return(asset_manager_response.deep_stringify_keys)
```
brucebolt: I think this stub is only needed because of the same issue I had before adding 6010c2b09840566dd1f36f428219a85535fb1ac5.  Asset Manager will still be requested, even if the Accept header isn't allowed.

spec/requests/assets_spec.rb
```diff
@@ -24,6 +24,15 @@
   end
   let(:parsed_response) { JSON.parse(response.body).deep_symbolize_keys }
 
+  shared_examples "passes request params to Asset Manager" do
+    it "preserves the request params" do
```
brucebolt: Nitpick: is "preserves the request params" the right wording for this test?  As the controller doesn't preserve the `file` in exactly the same format as the original request.

spec/requests/assets_spec.rb
```diff
@@ -24,6 +24,15 @@
...
+      expected_params = request_params.fetch(:asset).dup
+      expected_params[:file] = anything if expected_params.key?(:file)
```
brucebolt: Is it possible to use `tempfile` rather than `anything` when a file is provided, as we're not testing that transformation at all?

spec/requests/assets_spec.rb
```diff
@@ -24,6 +24,15 @@
...
+
+      expect(Services.asset_manager).to have_received(:update_asset).with(asset_id, hash_including(expected_params))
```
brucebolt: The `hash_including` means we could be sending extra parameters to Asset Manager, which maybe we don't want (e.g. if the `compact` were to be removed from the controller, we could see unexpected values for other params being sent to Asset Manager).  Maybe just test the entire hash is sent through?

brucebolt: Related to this, I don't think there's any test that the `auth_bypass_ids` are being sent to Asset Manager.

spec/requests/assets_spec.rb
```diff
@@ -358,6 +371,197 @@
...
+          before do
+            allow(SecureRandom).to receive(:uuid).and_return("token")
+            travel_to Time.zone.local(2026, 1, 1, 0, 0, 1)
+
+            subject
+          end
+
+          before do
+            subject
+          end
```
brucebolt: This isn't needed, as `subject` is already being called in the `before` block on line 436.

brucebolt (commented):
Thanks for breaking this up into smaller commits, it made it much easier to review 🙇.

I've left a few comments, but otherwise this looks like it'll almost be ready to approve.

spec/requests/assets_spec.rb
callumknights: This is problematic as passing the tempfile through the controller mutates it into a different Rack tempfile object, I'll see if there is a better way to do this.

spec/requests/assets_spec.rb
callumknights: This has snuck in by accident, removed 👍

spec/requests/assets_spec.rb
callumknights: I've added both of these changes to that shared_example. I've actually added them as part of the final commit, as this involves testing the `token` value, which is altered in the final commit.

spec/requests/assets_spec.rb
callumknights: Hopefully the reword I've added is okay

spec/requests/assets_spec.rb
callumknights: Should be sorted now

spec/requests/assets_spec.rb
callumknights: Thanks! I hadn't noticed this, I've removed the stub and moved the asset_manager call

spec/requests/assets_spec.rb
```diff
@@ -23,13 +23,29 @@
...
+      expect(decoded_token_payload_from_url(parsed_response[:file_url])).to eq(expected_decoded_token)
+    end
+
+    it "includes a preview expiry date 30 days in the future" do
+      expect(parsed_response).to include(preview_expiry: Time.zone.local(2026, 1, 31, 0, 0, 1).iso8601)
+    end
+  end
 
   shared_examples "passes request params to Asset Manager" do
     it "passes forward request params" do
```
brucebolt: I still think the naming of this test isn't really matching what's being tested, as the request params are not being passed to Asset Manager.

brucebolt (approved):
This looks good to me.  I've just left one non-blocking comment about a test name.

We should probably also apply the "passes request params to Asset Manager" tests to all the other endpoints, but that's out of scope for this PR.

spec/requests/assets_spec.rb
callumknights: Oh I see your meaning, as it's actually passing them to the `Services.asset_manager` stub. Would being clear about the stub in the test naming suffice?

brucebolt: My comment is less about where the params are being sent to, but that they aren't the same params from the client's original request.

Maybe it could be worded `passes the correct parameters to Asset Manager`?

callumknights: Thanks for explaining 🙇 I'll change it now

callumknights: This is now done, thanks again for explaining the issue.

brucebolt (approved):

